PR #218, opened 2024-12-03
Skip creating IRA resources in e2e on region that does not support IRA

2ez4szliu:
Some region (ap-southeast-5 for example) does not support IAM Roles Anywhere. We will need to skip creating IRA resources when setting up e2e test on those region.

We need to use a template with condition to only create IRA resource when the region supports IRA because it will cause template parsing error in those region if IRA resource exists in template.

`Unrecognized resource types: [AWS::RolesAnywhere::Profile, AWS::RolesAnywhere::TrustAnchor]`

### DIFF
--- a/test/e2e/stack.go
+++ b/test/e2e/stack.go
@@ -4,9 +4,12 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"fmt"
+	"slices"
+	"text/template"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -18,7 +21,9 @@ import (
 )
 
 //go:embed cfn-templates/hybrid-cfn.yaml
-var cfnTemplateBody []byte
+var cfnTemplateBody string
+
+var iraUnsupportedRegions = []string{"ap-southeast-5"}
 
 type e2eCfnStack struct {
 	clusterName            string
@@ -27,6 +32,10 @@ type e2eCfnStack struct {
 	cfn                    *cloudformation.CloudFormation
 	iam                    *iam.IAM
 	iamRolesAnywhereCACert []byte
+}
+
+type hybridCfnTemplateVars struct {
+	IncludeRolesAnywhere bool
 }
 
 type e2eCfnStackOutput struct {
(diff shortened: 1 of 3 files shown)


test/e2e/cfn-templates/hybrid-cfn-ssm-only.yaml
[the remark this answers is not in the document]
2ez4szliu: Done. We will use {{- if .IRASupported}} condition in the template

test/e2e/nodeadm_test.go
```diff
@@ -532,3 +537,7 @@ func removeSpecialChars(input string) string {
 	re := regexp.MustCompile(`[^a-zA-Z0-9-]+`)
 	return re.ReplaceAllString(input, "")
 }
+
+func isIraSupported(region string) bool {
```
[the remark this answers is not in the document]
2ez4szliu: makes sense. Move this logic into stack

test/e2e/stack.go
```diff
@@ -27,6 +32,11 @@ type e2eCfnStack struct {
 	cfn                    *cloudformation.CloudFormation
 	iam                    *iam.IAM
 	iamRolesAnywhereCACert []byte
+	region                 string
```
g-gaston, line 35: can't you get this from `e.cfn.ResolvedRegion`?

test/e2e/stack.go
```diff
@@ -89,12 +99,17 @@ func (e *e2eCfnStack) deployStack(ctx context.Context, logger logr.Logger) error
 			ParameterValue: aws.String(string(e.iamRolesAnywhereCACert)),
 		},
 	}
-
+	var buf bytes.Buffer
+	cfnTemplate := template.Must(template.New("hybridCfnTemplate").Parse(cfnTemplateBody))
```
g-gaston, line 103: is there a reason to use `must` instead of checking for an error and returning it?

test/e2e/stack.go
```diff
@@ -27,6 +32,11 @@ type e2eCfnStack struct {
 	cfn                    *cloudformation.CloudFormation
 	iam                    *iam.IAM
 	iamRolesAnywhereCACert []byte
+	region                 string
+}
+
+type hybridCfnTemplateVars struct {
+	IRASupported bool
```
g-gaston, line 39: nit: avoid non standard acronyms when possible. Also, naming things after what they do instead of what they are use for tends to draw better boundaries between components
```suggestion
	IncludeRolesAnywhere bool
```

test/e2e/stack.go
2ez4szliu: We will need to do `*e.cfn.Config.Region`

test/e2e/stack.go
2ez4szliu: Thanks, I think creating template without must and handle the error manually makes more sense. Removed Must

test/e2e/stack.go
2ez4szliu: Done

test/e2e/stack.go
```diff
@@ -90,11 +99,21 @@ func (e *e2eCfnStack) deployStack(ctx context.Context, logger logr.Logger) error
 		},
 	}
 
+	var buf bytes.Buffer
+	cfnTemplate, err := template.New("hybridCfnTemplate").Parse(cfnTemplateBody)
+	if err != nil {
+		return fmt.Errorf("Failed to parse hybrid-cfn.yaml template: %v", err)
```
g-gaston, line 105: nit: avoid using words like failed or error in an error message, it's redundant into. also error message should not be capitalized. also use %w for error wrapping
```suggestion
		return fmt.Errorf("parsing hybrid-cfn.yaml template: %w", err)
```

test/e2e/stack.go
```diff
@@ -90,11 +99,21 @@ func (e *e2eCfnStack) deployStack(ctx context.Context, logger logr.Logger) error
...
+	}
+	cfnTemplateConfig := &hybridCfnTemplateVars{IncludeRolesAnywhere: isIraSupported(*e.cfn.Config.Region)}
+	err = cfnTemplate.Execute(&buf, cfnTemplateConfig)
+	if err != nil {
+		return fmt.Errorf("Failed to apply data to hybrid-cfn.yaml template: %v", err)
```
g-gaston, line 110:
```suggestion
		return fmt.Errorf("applying data to hybrid-cfn.yaml template: %w", err)
```

g-gaston (approved):

PR #221, opened 2024-12-05
Move e2e code to separate tests from library

g-gaston:
## Description of changes
For the most part this is just moving code around. I had to create a couple new structs to avoid bad dependency relationships, but the abstractions remain the same (for better or for worse).

I intend to go deeper, better defining the purpose of each package, trimming down the public interface and in general making the boundaries more obvious.

This is just the first step, separating everything in packages that more or less makes sense, removes the assumption we had until now where everything was in a big messy context. I might end up changing/adding/removing some of this packages but at least for now they are good enough. The only part I'm not super happy with it the `e2e` package, that I'm basically using to dump everything that needs to be shared. I have a couple ideas I will try.

However, I ask to merge this as is to prevent more merge conflicts. Both on other PRs (if I make this bigger) and on mine when other PRs merge.

*Testing (if applicable):*
Ran this manually through one ssm and one iam-ra tests.


test/e2e/credentials/stack.go
```diff
@@ -216,19 +214,19 @@ func (s *e2eCfnStack) createInstanceProfile(ctx context.Context, logger logr.Log
...
 
-func (s *e2eCfnStack) instanceProfileName(roleName string) string {
+func (s *Stack) instanceProfileName(roleName string) string {
 	return roleName
 }
 
-func (e *e2eCfnStack) readStackOutput(ctx context.Context, logger logr.Logger) (*e2eCfnStackOutput, error) {
-	resp, err := e.cfn.DescribeStacksWithContext(ctx, &cloudformation.DescribeStacksInput{
-		StackName: aws.String(e.stackName),
+func (e *Stack) readStackOutput(ctx context.Context, logger logr.Logger) (*StackOutput, error) {
```
2ez4szliu, line 221: Now that we have removed e2e from those struct's name, do we want to rename this variable `e` to something else?

2ez4szliu (approved):

PR #222, opened 2024-12-05
Set ec2 instance type based on region and arch

2ez4szliu:
Create a method that returns the EC2 instance type that will be used in e2e test based region and OS architecture and remove the InstanceType definition in all OS packages.

*Testing (if applicable):*
Built e2e.test binary and ran test on a cluster on ap-southeast-5 region, an EC2 of type m6i.large was created and e2e test passed.

### DIFF
--- a/test/e2e/suite/nodeadm_test.go
+++ b/test/e2e/suite/nodeadm_test.go
@@ -391,7 +391,7 @@ var _ = Describe("Hybrid Nodes", func() {
 								ClusterName:        test.cluster.Name,
 								InstanceName:       instanceName,
 								AmiID:              amiId,
-								InstanceType:       os.InstanceType(),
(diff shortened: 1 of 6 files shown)


test/e2e/suite/nodeadm_test.go
```diff
@@ -555,3 +555,13 @@ func newE2EAWSSession(region string) (*session.Session, error) {
 	}
 	return sess, nil
 }
+
+func instanceType(region string, arch string) string {
```
g-gaston, line 559: Why can't the OSs just call this method? We are now exposing more details, making the interface more complicated

2ez4szliu: I thought the method Arch() might be better to be set on an interface called `NodeadmOS` than `InstanceType`. I might be wrong about this, anyway I have fixed this.

g-gaston (approved):

PR #223, opened 2024-12-06
Refactor e2e suite and tests setup

g-gaston:
## Description of changes
Next refactor piece after separating test code from "library". This PR focuses on the tests setup code, making some pieces reusable when possible and when not, splitting the long setup functions into smaller logic units. Most of that code is boiler plate, so by just moving it to separate functions with meaningful names we make reading the test code top to bottom easier.

In addition, a couple minor improvements/cleanups, that don't really change the flow of the test:
* Simplify the code that reads cluster details for a Hybrid cluster. Now it can infer all the data from just the cluster name by querying AWS APIs.
* Moved all that code to use go SDK v2. Eventually we should move all of it, but it's ok to do it piece meal.
* Now the `skipCleanup` is a flag that gets passed from the suite config. This toggle is still controller through the same env variable, but now it's only the suite setup code that reads it, and the rest of the test code just pulls this info from the the test config objects. This ensures the code behaves consistently for the duration of the test and allows to easily add/change the logic that drives the toggle.

### DIFF
--- a/test/e2e/suite/nodeadm_test.go
+++ b/test/e2e/suite/nodeadm_test.go
@@ -64,6 +64,7 @@ type TestConfig struct {
 
 type suiteConfiguration struct {
 	TestConfig             *TestConfig              `json:"testConfig"`
+	SkipCleanup            bool                     `json:"skipCleanup"`
 	CredentialsStackOutput *credentials.StackOutput `json:"ec2StackOutput"`
 	RolesAnywhereCACertPEM []byte                   `json:"rolesAnywhereCACertPEM"`
 	RolesAnywhereCAKeyPEM  []byte                   `json:"rolesAnywhereCAPrivateKeyPEM"`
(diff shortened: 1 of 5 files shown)


test/e2e/suite/nodeadm_test.go
```diff
@@ -539,19 +439,85 @@ func (u uninstallNodeTest) Run(ctx context.Context) error {
...
+// readTestConfig reads the configuration from the specified file path and unmarshals it into the TestConfig struct.
+func readTestConfig(configPath string) (*TestConfig, error) {
+	config := &TestConfig{}
+	file, err := os.ReadFile(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("reading tests configuration file %s: %w", filePath, err)
+	}
+
+	if err = yaml.Unmarshal(file, config); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal configuration from YAML: %v", err)
```
2ez4szliu, line 451:
```suggestion
		return nil, fmt.Errorf("unmarshalling configuration from YAML: %w", err)
```

2ez4szliu (approved):

PR #224, opened 2024-12-09
Create E2E setup using cloudformation stack and aws-go-sdk-v2 apis

panktishah26:
*Description of changes:*
Create E2E setup using cloudformation stack and aws-go-sdk-v2 APIs.

*Testing (if applicable):*


hybrid-nodes-cdk/lib/nodeadm-stack.ts
```diff
@@ -218,7 +220,6 @@ export class NodeadmBuildStack extends cdk.Stack {
           actions: [
             'iam:DeleteRolePolicy',
             'iam:DetachRolePolicy',
```
abhay-krishna, line 222: Do we still need the `DetachRolePolicy` permission here if we're adding it in the block above?

g-gaston (approved):

PR #228, opened 2024-12-10
Fix instance type assignment based on architecture

abhay-krishna:
This PR fixes the instance type assignment based on architecture. Earlier we were only checking if the architecture was `x86_64` to assign an AMD64 instance type, and otherwise assigning ARM64 if the check fails, but Ubuntu uses `amd64` and so the check was failing, leading to us using an ARM64 instance type for an AMD64 AMI which led to the error
```console
The architecture 'arm64' of the specified instance type does not match the architecture 'x86_64' of the specified AMI. Specify an instance type and an AMI that have matching architectures, and try again.
```
Now we are checking for both `amd64` and `x86_64` when assigning instance types.

### DIFF
--- a/test/e2e/os/arch.go
+++ b/test/e2e/os/arch.go
@@ -11,8 +11,9 @@ import (
 )
 
 const (
-	amd64Arch = "x86_64"
+	amd64Arch = "amd64"
 	arm64Arch = "arm64"
+	x8664Arch = "x86_64"
 )
 
 func populateBaseScripts(userDataInput *e2e.UserDataInput) error {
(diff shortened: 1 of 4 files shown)


test/e2e/os/arch.go
```diff
@@ -58,7 +59,7 @@ func getAmiIDFromSSM(ctx context.Context, client *ssm.SSM, amiName string) (*str
 }
 
 func getInstanceTypeFromRegionAndArch(region, arch string) string {
-	if arch == amd64Arch {
+	if arch == amd64Arch || arch == x8664Arch {
```
g-gaston, line 62: can we just normalize the arch string before passing it to this method?

abhay-krishna: But that's just going to be another method with a similar control statement just the arch. I didn't do that to avoid the duplication.

g-gaston: Why is another method bad? 

I like the idea bc it would encapsulate the logic to determine the architecture in each os, making this code decoupled from the per os string values. It also allows you to describe some of the business logic in the type system by using an "enum".

abhay-krishna: Not bad per se, just went with what IMO was the easiest solution for the short term, but I have updated it to pass in the normalized arch.

2ez4szliu (approved):

PR #232, opened 2024-12-11
adds upgrade e2e test

jaxesn:
*Issue #, if available:*

*Description of changes:*

This builds on #189 adding the upgrade tests to the e2e suite.  Nodes are created as n-1 version worker nodes and then after joining, are upgraded and the node name is confirmed to not change while the version is bumped.

These are added as new ginkgo table tests for all versions/OSs, but in our e2e script we are limiting how many run until we can better limit our vcpu/instance usage.

- Pulls in kubectl to use the cordon/drain functionality
- Added draining to the existing uninstall flow
- Overrides both the cilium and calico tolerations so that the operators/controllers do not tolerate cordoned nodes. There are upstream issues for this in both repos. This allows us to run uninstall and upgrade without skipping the preflights
- Cleans up the ssm calls a bit to reuse the same base method.

*Testing (if applicable):*


test/e2e/suite/nodeadm_test.go
```diff
@@ -348,12 +283,182 @@ var _ = Describe("Hybrid Nodes", func() {
...
+	nodeadmURLs        e2e.NodeadmURLs
+	os                 e2e.NodeadmOS
+	provider           e2e.NodeadmCredentialsProvider
+	s3Client           *s3v1.S3
+	skipCleanup        bool
+	ssmClient          *ssmv1.SSM
+}
+
+func (c createNodeTest) Run(ctx context.Context) (ec2.Instance, error) {
+	Expect(c.os).NotTo(BeNil())
```
jaxesn, line 377: I can remove all these expects and just return the err if thats what we want?

test/e2e/suite/nodeadm_test.go
```diff
@@ -348,12 +283,182 @@ var _ = Describe("Hybrid Nodes", func() {
 						},
 						Entry(fmt.Sprintf("With OS %s and with Credential Provider %s", os.Name(), string(provider.Name())), context.Background(), os, provider, Label(os.Name(), string(provider.Name()), "simpleflow")),
 					)
+
+					DescribeTable("Upgrade nodeadm flow",
+						func(ctx context.Context, os e2e.NodeadmOS, provider e2e.NodeadmCredentialsProvider) {
+
+							// Skip upgrade flow for cluster with the minimum kubernetes version
+							if test.cluster.KubernetesVersion == kubernetes.MinimumVersion {
```
g-gaston, line 291: This should probably be a `<=` for correctness

test/e2e/suite/nodeadm_test.go
g-gaston: yeah if the method already returns an error, we should do that.

test/e2e/suite/nodeadm_test.go
```diff
@@ -348,12 +283,182 @@ var _ = Describe("Hybrid Nodes", func() {
...
+	Expect(err).NotTo(HaveOccurred())
+
+	nodeadmConfig, err := c.provider.NodeadmConfig(ctx, nodeSpec)
+	Expect(err).NotTo(HaveOccurred(), "expected to build nodeconfig")
+
+	nodeadmConfigYaml, err := yaml.Marshal(&nodeadmConfig)
+	Expect(err).NotTo(HaveOccurred(), "expected to successfully marshal nodeadm config to YAML")
+
+	var rootPasswordHash string
+	if suite.TestConfig.SetRootPassword {
```
g-gaston, line 399: can we pass this in the createNodeTest?
Ideally only the `BeforeEach` should be reading the `suite` variable (I think I missed this one before)
Same with the other `suite` references in this method

test/e2e/kubernetes/kubernetes.go
```diff
@@ -251,3 +257,99 @@ func EnsureNodeWithIPIsDeleted(ctx context.Context, k8s *kubernetes.Clientset, i
 	}
 	return nil
 }
+
+func WaitForHybridNodeToUpgrade(ctx context.Context, k8s *kubernetes.Clientset, nodeName, targetVersion string, logger logr.Logger) (*corev1.Node, error) {
```
g-gaston, line 261:
```suggestion
func WaitForNodeToHaveVersion(ctx context.Context, k8s *kubernetes.Clientset, nodeName, targetVersion string, logger logr.Logger) (*corev1.Node, error) {
```

test/e2e/kubernetes/kubernetes.go
```diff
@@ -251,3 +257,99 @@ func EnsureNodeWithIPIsDeleted(ctx context.Context, k8s *kubernetes.Clientset, i
...
+		return nil, fmt.Errorf("waiting for node %s kubernetes version to be upgraded to %s: %w", nodeName, targetVersion, err)
+	}
+
+	return foundNode, nil
+}
+
+func GetPreviousK8sVersion(kubernetesVersion string) (string, error) {
+	currentVersion, err := version.ParseSemantic(kubernetesVersion + ".0")
+	if err != nil {
+		return "", fmt.Errorf("parsing kubernetes version: %v", err)
```
g-gaston, line 297: nit:
```suggestion
		return "", fmt.Errorf("parsing version: %w", err)
```

test/e2e/kubernetes/kubernetes.go
```diff
@@ -251,3 +257,99 @@ func EnsureNodeWithIPIsDeleted(ctx context.Context, k8s *kubernetes.Clientset, i
...
+		return false, nil // continue polling
+	})
+	if err != nil {
+		return nil, fmt.Errorf("waiting for node %s kubernetes version to be upgraded to %s: %w", nodeName, targetVersion, err)
+	}
+
+	return foundNode, nil
+}
+
+func GetPreviousK8sVersion(kubernetesVersion string) (string, error) {
```
g-gaston, line 294: Avoid using `get` unless you are actually calling a get API. This is not a getter but same argument applies imo: https://go.dev/doc/effective_go#Getters

```suggestion
func PreviousVersion(version string) (string, error) {
```

test/e2e/suite/nodeadm_test.go
```diff
@@ -348,12 +291,204 @@ var _ = Describe("Hybrid Nodes", func() {
...
+	}
+
+	nodeadmConfig, err := c.provider.NodeadmConfig(ctx, nodeSpec)
+	if err != nil {
+		return ec2.Instance{}, fmt.Errorf("expected to build nodeconfig: %v", err)
+	}
+
+	nodeadmConfigYaml, err := yaml.Marshal(&nodeadmConfig)
+	if err != nil {
+		return ec2.Instance{}, fmt.Errorf("expected to successfully marshal nodeadm config to YAML: %v", err)
```
g-gaston, line 415: prefer to use %w instead of %v for errors. %w wraps the error, so things like `Unwrap`/`As`/`Is` work properly
Sometimes you might not want to wrap, like if you want to hide the underlying type and not allow to check for it up the chain. But in most cases wrap is what you want

```suggestion
		return ec2.Instance{}, fmt.Errorf("expected to successfully marshal nodeadm config to YAML: %w", err)
```

g-gaston (approved):

g-gaston (approved):

g-gaston (approved):

PR #233, opened 2024-12-11
Normalize OS architecture

g-gaston:
*Description of changes:*
We have probably overgrew the original architecture abstraction and we end ended up using it for multiple things. The differences between the ami arch name between OS made the code prone to errors and required having a lot of context to modify safely.

### DIFF
--- a/test/e2e/os/amazonlinux.go
+++ b/test/e2e/os/amazonlinux.go
@@ -19,34 +19,34 @@ type amazonLinuxCloudInitData struct {
 }
 
 type AmazonLinux2023 struct {
-	Architecture string
+	amiArchitecture string
+	architecture    architecture
 }
 
 func NewAmazonLinux2023AMD() *AmazonLinux2023 {
 	al := new(AmazonLinux2023)
-	al.Architecture = x8664Arch
+	al.amiArchitecture = x8664Arch
+	al.architecture = amd64
 	return al
 }
 
 func NewAmazonLinux2023ARM() *AmazonLinux2023 {
 	al := new(AmazonLinux2023)
-	al.Architecture = arm64Arch
+	al.amiArchitecture = arm64Arch
+	al.architecture = arm64
 	return al
 }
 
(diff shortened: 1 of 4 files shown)


test/e2e/os/arch.go
```diff
@@ -58,15 +74,8 @@ func getAmiIDFromSSM(ctx context.Context, client *ssm.SSM, amiName string) (*str
...
-	if arch == x8664Arch {
-		return amd64Arch
-	}
-	return arch
-}
-
-func getInstanceTypeFromRegionAndArch(region, arch string) string {
-	if arch == amd64Arch {
+func getInstanceTypeFromRegionAndArch(region string, arch architecture) string {
+	if arch.arm() {
```
abhay-krishna, line 78:
```suggestion
	if arch.amd() {
```

test/e2e/os/amazonlinux.go
```diff
@@ -19,34 +19,34 @@ type amazonLinuxCloudInitData struct {
...
+	al.architecture = arm64
 	return al
 }
 
 func (a AmazonLinux2023) Name() string {
-	if a.Architecture == x8664Arch {
-		return "al23-amd64"
-	}
-	return "al23-arm64"
+	return "al23-" + a.architecture.String()
```
abhay-krishna, line 41: Just a nit that I missed during previous reviews, but any reason why we shouldn't change all of these to `Sprintf`s? Or maybe even a `generateName` variadic function whose arguments you can append to the OS name.

g-gaston: For simple string concatenation, `+` is usually better (IMO): more expressive, simpler and easier to read. As a bonus, it's also *faster*, but tbh that's not a concern here.

About having a function for name generation, I can see that if we want to standarize the OS name format. However, I'm not sure that's what we nee here. In fact, they are already not standard since some of them have the `-docker` suffix. I'm more inclined to leave it as it is.

abhay-krishna (approved):

PR #246, opened 2024-12-17
improves swap partition checking

jaxesn:
*Issue #, if available:*

*Description of changes:*

If nodeadm detects a swap partition, it errors out to the user to disable that.  If its a file, it can remove/disable it.  The logic that checked for this was exec swapon/awk.  This changes the code to be more like kubeadm's preflight check.  Added a new integration test to validate this.

Improved integration tests docker setup:
- fixed the dockerignore location/paths
- use public ecr golang image from eksd
- pull mock metadata built artifact instead of go install
- copy go.mod/sum and run download to improve rerun caching

*Testing (if applicable):*

### DIFF
--- a/internal/system/swap.go
+++ b/internal/system/swap.go
@@ -56,14 +56,25 @@ func (s *swapAspect) Setup() error {
 // If partition type swap exists, user needs to manually remove the partition swap before
 // running nodeadm init.
 func partitionSwapExists() (bool, error) {
-	cmd := "swapon -s | awk '$2==\"partition\" {print}'"
-	out, err := exec.Command("bash", "-c", cmd).CombinedOutput()
+	f, err := os.Open("/proc/swaps")
 	if err != nil {
-		return false, fmt.Errorf("failed to check if partition type swap exists: %v", err)
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("checking if partition type swap exists: %v", err)
 	}
-	if len(string(out)) != 0 {
-		return true, nil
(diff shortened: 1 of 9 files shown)


internal/system/swap.go
```diff
@@ -56,14 +56,22 @@ func (s *swapAspect) Setup() error {
 // If partition type swap exists, user needs to manually remove the partition swap before
 // running nodeadm init.
 func partitionSwapExists() (bool, error) {
-	cmd := "swapon -s | awk '$2==\"partition\" {print}'"
-	out, err := exec.Command("bash", "-c", cmd).CombinedOutput()
+	f, err := os.Open("/proc/swaps")
 	if err != nil {
```
g-gaston, line 60: should we check here if the error is file not found?
IMO there are two cases: the file doesn't exist and the file exists but we got an error when opening it

internal/system/swap.go
```diff
@@ -56,14 +56,22 @@ func (s *swapAspect) Setup() error {
...
+	for scanner.Scan() {
+		fields := strings.Fields(scanner.Text())
+		if fields[1] == "partition" {
+			return true, nil
+		}
 	}
-	if len(string(out)) != 0 {
-		return true, nil
+	if err := scanner.Err(); err != nil {
+		return false, fmt.Errorf("failed to check if partition type swap exists: %v", err)
```
g-gaston, line 72:
```suggestion
		return false, fmt.Errorf("checking if partition type swap exists: %w", err)
```

g-gaston (approved):

PR #247, opened 2024-12-17
Added a SecurityGroup for RemoteNode and RemotePod

raykrueger:
An EKS Cluster supporting Hybrid Nodes requires a security group that allows port 443 from those cidrs. This commit adds a secondary security group for those cidrs.

This replaces the SecurityGroupId parameter with a required VpcId that the SecurityGroup will be created in.

### DIFF
--- a/example/hybrid-eks-cfn.yaml
+++ b/example/hybrid-eks-cfn.yaml
@@ -11,6 +11,10 @@ Parameters:
   ClusterRoleName:
     Type: String
     Default: 'EKSClusterRole'
+  VpcId:
+    Type: String
+    Description: The VPC that a Security Group will be created in for the EKS Cluster, required if a SecurityGroupId is not configured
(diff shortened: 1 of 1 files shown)


csplinter:
We need to update the docs with this change ([this page](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-cluster-create.html#hybrid-nodes-cluster-create-cfn))

In the docs, we need to say that the CloudFormation template creates the additional security group with the ingress for the remote node and pod CIDRs.

raykrueger:
The current configuration matches the EKS Control plane ENI default security group outbound rules.

csplinter:
@finchd can we update the docs page linked below with the change required with this CloudFormation template change?

csplinter (approved):
Validated with and without security group ID, looks good

PR #249, opened 2024-12-17
adds support to e2e for ecr repos with pull through cache

jaxesn:
*Issue #, if available:*

*Description of changes:*

Changes cni image pulling to come from newly setup ecr repos with pull through cache setup so we do not have to manually populate these when bumping cilium/calico versions.

The mirror-* scripts have been change to do just a image inspect, which will trigger the pull through if the image does not already exist.  These scripts are now run during the CI build so that the caches are populated before tests.

*Testing (if applicable):*


test/e2e/cluster/create.go
```diff
@@ -15,12 +15,14 @@ import (
...
-	ClusterNetwork    NetworkConfig `yaml:"clusterNetwork"`
-	HybridNetwork     NetworkConfig `yaml:"hybridNetwork"`
-	KubernetesVersion string        `yaml:"kubernetesVersion"`
-	Cni               string        `yaml:"cni"`
+	ClusterName       string           `yaml:"clusterName"`
+	ClusterRegion     string           `yaml:"clusterRegion"`
+	ClusterNetwork    NetworkConfig    `yaml:"clusterNetwork"`
+	HybridNetwork     NetworkConfig    `yaml:"hybridNetwork"`
+	KubernetesVersion string           `yaml:"kubernetesVersion"`
+	Cni               string           `yaml:"cni"`
```
jaxesn, line 23: i dont love what i did here, looking for feedback/other options?  One I thought migth be better than this would be to turn the cni field into the config instead of a string?

As background, the slightly complicating bit here is for calico we need to give it two registry/paths. For ecr when you setup the ptc it has to be namespaced, so `<accountID>.<ecr>.com/docker.io` or `<accountID>.<ecr>.com/quay.io`. In the case of calico, the operator is in quay and the rest are from docker.

jaxesn: Here is an example config

```
clusterName: nodeadm-e2e-tests-fc7ea9db-733d-445b-9af5-82fcf1497092
clusterRegion: us-west-2
kubernetesVersion: 1.31
cni: cilium
calico:
  containerRegistry: <acct>.dkr.ecr.us-west-2.amazonaws.com/docker.io
  operatorContainerRegistry: <acct>.dkr.ecr.us-west-2.amazonaws.com/quay.io
cilium:
  containerRegistry: <acct>.dkr.ecr.us-west-2.amazonaws.com/quay.io
clusterNetwork:
  vpcCidr: 10.0.0.0/16
  publicSubnetCidr: 10.0.10.0/24
  privateSubnetCidr: 10.0.20.0/24
hybridNetwork:
  vpcCidr: 10.1.0.0/16
  publicSubnetCidr: 10.1.1.0/24
  privateSubnetCidr: 10.1.2.0/24
  podCidr: 10.2.0.0/16
```

jaxesn: Thinking more, if we are willing to break from the current ecr we use today and standardize on `quay.io` `docker.io`, potentially even `ecr-public`, as the namespaces we could just add `containerRegistry` at the root of the spec, not specific to the cni, and just hardcode the namespace/prefixes accordingly.   We could even handle the nginx container this way as well.

test/e2e/cni/testdata/calico/mirror-calico.sh
```diff
@@ -15,30 +15,21 @@
...
-	dst=${DST_REGISTRY}/${repo_with_version}
-	oras cp ${image} ${dst}
-done
+	for image in $(curl -s -L https://github.com/projectcalico/calico/releases/download/$CALICO_VERSION/metadata.yaml | yq  -r ".images[]" ); do
+		repo_with_version=$image
+		if [[ $image != quay* ]]; then
+			repo_with_version="docker.io/$image"
+		fi
+		
+		docker buildx imagetools inspect "${registry}/${repo_with_version}"
```
g-gaston, line 33: could we add a comment here explaining that this does a head (or probably a get of the manifest?), which triggers the pull through in the ecr repo to populate the cache? If I didn't have context about this I think it wouldn't be obvious to me

same in the cilium script if you can

buildspecs/build-nodeadm.yml
```diff
@@ -8,6 +8,9 @@ phases:
     - aws s3 cp _bin/arm64/nodeadm s3://$ARTIFACTS_BUCKET/latest-pre/linux/arm64/nodeadm
     - aws s3 cp _bin/e2e-test s3://$ARTIFACTS_BUCKET/latest-pre/e2e-test
     - aws s3 cp _bin/e2e.test s3://$ARTIFACTS_BUCKET/latest-pre/e2e.test
+    - mkdir -p /root/.docker && cp hack/docker-ecr-config.json /root/.docker/config.json
+    - ./test/e2e/cni/testdata/cilium/mirror-cilium.sh
+    - ./test/e2e/cni/testdata/calico/mirror-calico.sh
```
g-gaston, line 13: What would you think about putting this in a different stage? It could even run in parallel to the build one since there is no dependency. Although the main benefit to me is separation of concerns (pre-populating the cache is not part of building the binaries)

g-gaston (approved):

PR #255, opened 2024-12-23
adds a jumpbox to e2e tests

jaxesn:
*Issue #, if available:*

*Description of changes:*

Currently we use SSM to run remote commands on the worker nodes, since we are using EC2 instances.  This creates a kind of odd dependency which does not properly mirror on prem infra.  It also introduces some potential funky failure cases due to SSM on the cloud node vs SSM on the "hybrid" node.

This change introduces a jump box to the setup stack to be used for running these remote commands.  The tests now use SSM to the jump box which uses ssh to run the actual remote command on the worker node.  

I have tested by running the full suite of tests in my account many times and have seen really good success.  I believe this change has dramatically improved the success rate of the rhel8 tests, which used to have random failures due to disk corruption, which I think was due to how we used to reboot the instances.

*Testing (if applicable):*


test/e2e/credentials/setup.go
```diff
@@ -55,11 +62,46 @@ func Setup(ctx context.Context, logger logr.Logger, awsSession *session.Session,
...
+				Name:   aws.String("instance-state-name"),
+				Values: []string{"running"},
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	keypair, err := ec2Client.DescribeKeyPairs(ctx, &ec2.DescribeKeyPairsInput{
```
jaxesn, line 85: This is really the only reason we need the keypair, to get the public key. The key pair is never used in an AWS sense, attached to a node.  Its somewhat handy to "store" it in aws, but we could instead just read the pub key from the local file and/or config.

test/e2e/ssm/command.go
```diff
@@ -34,12 +34,17 @@ func (s *ssmConfig) runCommandsOnInstance(ctx context.Context, logger logr.Logge
...
+		return nil, fmt.Errorf("got an error calling SendCommand: %w", err)
+	}
 	invocationInput := &ssm.GetCommandInvocationInput{
 		CommandId:  output.Command.CommandId,
 		InstanceId: aws.String(s.instanceID),
 	}
 	// Will wait on Pending, InProgress, or Cancelling until we reach a terminal status of Success, Cancelled, Failed, TimedOut
-	_ = s.client.WaitUntilCommandExecutedWithContext(ctx, invocationInput)
+	_ = s.client.WaitUntilCommandExecutedWithContext(ctx, invocationInput, func(w *request.Waiter) {
+		w.MaxAttempts = 60
```
jaxesn, line 46: ill add a code comment, this is because upgrade can take longer than the default timeout/retries, esp on rhel.

test/e2e/cluster/cfn-templates/setup-cfn.yaml
```diff
@@ -290,6 +298,64 @@ Resources:
...
+        - Key: !Ref TestClusterTagKey
+          Value: !Ref ClusterName
+
+  JumpboxInstanceProfile:
+    Type: AWS::IAM::InstanceProfile
+    Properties:
+      Roles: 
+        - !Ref JumpboxInstanceRole
+
+  JumpboxKeyPair:
```
jaxesn, line 323: see note below about this keypair.

hack/e2e-cleanup.sh
```diff
@@ -177,6 +177,19 @@ function iam_ra_cluster_name_tag_from_resource(){
     echo "$cluster_name"
 }
 
+# See above note about role tags
+function ssm_parameter_cluster_name_tag(){
+    id="$1"
+    >&2 echo "($(pwd)) \$ command aws ssm list-tags-for-resource --resource-id $id --resource-type Parameter --query "{Tags:TagList}" --output json"
```
g-gaston, line 183: remove?

test/e2e/suite/nodeadm_test.go
```diff
@@ -704,16 +719,18 @@ func buildPeeredVPCTestForSuite(ctx context.Context, suite *suiteConfiguration)
...
-	logger   logr.Logger
-	provider e2e.NodeadmCredentialsProvider
-	ssm      *ssmv1.SSM
+	cluster             *peered.HybridCluster
+	ec2                 ec2.Instance
+	k8s                 *clientgo.Clientset
+	logger              logr.Logger
+	provider            e2e.NodeadmCredentialsProvider
+	remoteCommandRunner commands.RemoteCommandRunner
+	ssm                 *ssmv1.SSM
```
g-gaston, line 733: we can prob remove ssm and provider?

test/e2e/suite/nodeadm_test.go
```diff
@@ -563,11 +580,12 @@ func (t joinNodeTest) Run(ctx context.Context) error {
 }
 
 type uninstallNodeTest struct {
-	k8s      *clientgo.Clientset
-	ssm      *ssmv1.SSM
-	ec2      ec2.Instance
-	provider e2e.NodeadmCredentialsProvider
-	logger   logr.Logger
+	k8s                 *clientgo.Clientset
+	ssm                 *ssmv1.SSM
```
g-gaston, line 584: we can prob remove ssm

hack/e2e-cleanup.sh
jaxesn: ive been keeping these in this script to make it easier to track whats going on, if you are trying to follow the logs.  The reason not all calls all this is that most of the calls `aws` is wrapped by `echo_and_run`, but for these where we check the error code, they are not so I have this manual log message.

abhay-krishna (approved):

PR #259, opened 2024-12-28
improves instance creation retry

jaxesn:
*Issue #, if available:*

*Description of changes:*

When running all the tests concurrently there would usually be a few that would fail to create the instance, even with the 60 retries.  This removes the wrapping of the standard retrier in favor of using more of the defaults and passing a custom iserrorretryable.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/ec2/instance.go
+++ b/test/e2e/ec2/instance.go
@@ -6,8 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"time"
 
+	awsV2 "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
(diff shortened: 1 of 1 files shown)


test/e2e/ec2/instance.go
```diff
@@ -135,24 +138,12 @@ func DeleteEC2Instance(ctx context.Context, client *ec2.Client, instanceID strin
...
-	maxRetries int
-	backoff    time.Duration
+	return runInstanceRetrier{}
 }
 
-func (c runInstanceRetrier) IsErrorRetryable(err error) bool {
-	if c.Standard.IsErrorRetryable(err) {
-		return true
-	}
+type runInstanceRetrier struct{}
```
g-gaston, line 144: now that this is not a retrier, should we change the name?
like
```suggestion
type invalidInstanceProfileRetryable struct{}
```

g-gaston (approved):

drewvanstone (approved):

PR #261, opened 2024-12-30
waits for cordon to complete

jaxesn:
*Issue #, if available:*

*Description of changes:*

When running all the tests concurrently there was generally a failure in some tests due to nodes being cordoned and drained, but would error out during upgrade due to pods running on the node.  We think this is due to cordon taking some time to fully taint the node and with our tests constantly draining nodes, the cilium/calico operator will be rescheduled often and was triggering this failure.

This adds a wait to ensure the node has the proper taint before attempting to drain the node.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/kubernetes/kubernetes.go
+++ b/test/e2e/kubernetes/kubernetes.go
@@ -30,6 +30,8 @@ const (
 	hybridNodeWaitTimeout    = 10 * time.Minute
 	hybridNodeDelayInterval  = 5 * time.Second
 	hybridNodeUpgradeTimeout = 2 * time.Minute
+	nodeCordonDelayInterval  = 1 * time.Second
+	nodeCordonTimeout        = 30 * time.Second
 	MinimumVersion           = "1.25"
 )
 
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/kubernetes.go
```diff
@@ -367,9 +364,44 @@ func CordonNode(ctx context.Context, k8s *kubernetes.Clientset, node *corev1.Nod
 		return fmt.Errorf("cordoning node %s: %v", node.Name, err)
 	}
 
+	nodeName := node.ObjectMeta.Name
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, nodeCordonDelayInterval, nodeCordonTimeout, true, func(ctx context.Context) (done bool, err error) {
```
g-gaston, line 369: nit: can we add a comment here saying why we do the polling?

g-gaston (dismissed):

g-gaston (approved):

PR #263, opened 2025-01-02
retry running ssm install to avoid flaky install failures

jaxesn:
*Issue #, if available:*

*Description of changes:*

While the full tests suite, I have noticed a failure which occurs about every other run, for a test or two.  

It seems mostly related to rhel and I believe its specifically due to the `insights-client` which automatically runs on boot of a rhel node and will attempt to dnf/rpm install a few rhel specific packages.  

This install process appears to line up, sometimes, with our call to ssm install and ends up causing it to fail with

`error: can't create transaction lock on /var/lib/rpm/.rpm.lock (Resource temporarily unavailable)`

This changes adds retries to the ssm install, like we do for [containerd](https://github.com/aws/eks-hybrid/blob/33e2bb92815d748c66a084300a1297557cf2cf47/internal/containerd/install.go#L43) and [iptables](https://github.com/aws/eks-hybrid/blob/33e2bb92815d748c66a084300a1297557cf2cf47/internal/iptables/iptables.go#L28)

*Testing (if applicable):*

### DIFF
--- a/internal/ssm/install.go
+++ b/internal/ssm/install.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/config"
 	awsSsm "github.com/aws/aws-sdk-go-v2/service/ssm"
@@ -14,6 +15,7 @@ import (
 
 	"github.com/aws/eks-hybrid/internal/artifact"
 	"github.com/aws/eks-hybrid/internal/tracker"
+	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


internal/ssm/install.go
```diff
@@ -42,10 +44,9 @@ func Install(ctx context.Context, tracker *tracker.Tracker, source Source) error
 		return errors.Wrap(err, "failed to install ssm installer")
 	}
 
-	installCmd := exec.Command(installerPath, "-install", "-region", DefaultSsmInstallerRegion)
-	out, err := installCmd.CombinedOutput()
+	err = runInstallWithRetries(ctx)
 	if err != nil {
```
g-gaston, line 48: nit:
```suggestion
	if err = runInstallWithRetries(ctx); err != nil {
```

g-gaston (dismissed):

g-gaston (approved):

PR #275, opened 2025-01-11
refactor: validate containerd daemon on init, not install

TylerGillson:
*Issue #, if available:*

Unable to run `nodeadm install` within a docker container. D-Bus connection fails due to lack of fully operational systemd:
```
INFO {“level”:“fatal”,“ts”:1736533734.779155,“caller”:“./main.go:53”,“msg”:“Command failed”,“error”:“please install systemd unit file for containerd: dial unix /run/systemd/private: connect: no such file or directory”,“stacktrace”:“main.main\n\t./main.go:53\nruntime.main\n\truntime/proc.go:272”}
INFO The command ‘/bin/sh -c /opt/nodeadmutil/bin/nodeadm install -p iam-ra 1.31.0’ returned a non-zero code: 1
```

*Description of changes:*

Relocate containerd validation from `nodeadm install` to `nodeadm init`.

*Testing (if applicable):*

N/A


g-gaston (commented):
Thanks for the PR @TylerGillson!

I would like to understand more about the usecase. Why do you need to run `install` from a container?

TylerGillson:
> Thanks for the PR @TylerGillson!
> 
> I would like to understand more about the usecase. Why do you need to run `install` from a container?

Of course! We support EKS hybrid nodes using [kairos](https://kairos.io/), which uses an immutable filesystem. Upgrades are shipped as container images built from [CanvOS](https://github.com/spectrocloud/CanvOS). Building these assets requires executing `nodeadm install` inside a container.

jaxesn:
@TylerGillson thanks for the submission!  Seems like a super interesting use case, curious did you have success launching the container (once you got install to pass) and joining to the EKS cluster on the kairos node?

TylerGillson:
Hey @jaxesn. The container's filesystem is ultimately unpacked to the filesystem of an actual VM and yes, that VM boots and is able to successfully pair with the EKS control plane. I just want to clarify that we're not running containerized worker nodes.

jaxesn:
Oh cool, thanks for the explanation!

g-gaston:
@TylerGillson 

We would love to enable your usecase, thanks for the context!

However, I feel like the issue is not the validation itself, but when in the workflow we perform it. What if instead of validating this during `install`, we validated it at the beginning of `init`?

Not only we avoid the need for another skip toggle, but it also feels like a better place. `install` doesn't care about you having a systemd unit file or not, `init` does care because it will start/stop the service.

If you have the time to work on that we would love to take the change in. Let me know what you think :)

TylerGillson:
Makes perfect sense to me, @g-gaston. I split the docs changes out into https://github.com/aws/eks-hybrid/pull/301. Please take another look 🙏🏼

cmd/nodeadm/init/init.go
```diff
@@ -72,6 +74,10 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		} else if err != nil {
 			return err
 		}
+
+		if err := containerd.ValidateSystemdUnitFile(); err != nil {
+			return fmt.Errorf("please install systemd unit file for containerd: %v", err)
```
g-gaston, line 79: nit:
```suggestion
			return fmt.Errorf("a systemd unit file for containerd is required to init the node: %w", err)
```

g-gaston (commented):
looks good, thanks a lot! just a nit

I think you have some kind of messy commit history, once you fix that this looks ready to go :)

cmd/nodeadm/init/init.go
TylerGillson: I preserved the exact wording from before, but happy to update it.

TylerGillson:
> looks good, thanks a lot! just a nit
> 
> I think you have some kind of messy commit history, once you fix that this looks ready to go :)

That all goes away if you do a squash merge. Wasn't sure how you guys handle merges around here. I can fix it up.

TylerGillson:
Closed in favour of https://github.com/aws/eks-hybrid/pull/275.

PR #279, opened 2025-01-13
changes e2e log collection to happen only once

jaxesn:
*Issue #, if available:*

*Description of changes:*

Now that we are using the jumpbox to run remote commands on the worker nodes we can simplify the log collection.  

We used to rely on running the collection in traps in each "bash" script we ran since during some of our phases we may "break" the ec2 ssm, or in the case of cloud-init its just running during booth, to ensure we could capture as many logs, as often as we could.  Now, as long the node boots, we can rely on the jumpbox and ssh to be able to run the log collector.

This also removes the extract complexity around pre generating the links and using them by name, with a fallback.

Now the logs are only collected in the deferCleanup, on failure or success.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/suite/nodeadm_test.go
+++ b/test/e2e/suite/nodeadm_test.go
@@ -402,6 +402,10 @@ type createNodeTest struct {
 }
 
 func (c createNodeTest) Run(ctx context.Context) (ec2.Instance, error) {
+	if c.logsBucket != "" {
+		c.logger.Info(fmt.Sprintf("Logs bucket: https://%s.console.aws.amazon.com/s3/buckets/%s?prefix=%s/", c.cluster.Region, c.logsBucket, c.logsPrefix()))
+	}
+
 	nodeSpec := e2e.NodeSpec{
 		OS:         c.os,
 		NamePrefix: c.nodeNamePrefix,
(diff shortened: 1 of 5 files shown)


test/e2e/suite/nodeadm_test.go
```diff
@@ -501,6 +496,28 @@ func (c createNodeTest) Run(ctx context.Context) (ec2.Instance, error) {
...
+}
+
+func (c createNodeTest) collectLogs(ctx context.Context, bundleName, instanceIP string) error {
+	if c.logsBucket == "" {
+		return nil
+	}
+	key := fmt.Sprintf("%s/%s.tar.gz", c.logsPrefix(), bundleName)
+	url, err := s3.GeneratePutLogsPreSignedURL(ctx, c.s3Client, c.logsBucket, key, 5*time.Minute)
+	if err != nil {
+		c.logger.Info("collecting logs on node", err)
```
g-gaston, line 509: why log if we are returning the error?

test/e2e/suite/nodeadm_test.go
```diff
@@ -489,6 +479,11 @@ func (c createNodeTest) Run(ctx context.Context) (ec2.Instance, error) {
 	c.logger.Info(fmt.Sprintf("EC2 Instance Connect: https://%s.console.aws.amazon.com/ec2-instance-connect/ssh?connType=serial&instanceId=%s&region=%s&serialPort=0", c.cluster.Region, instance.ID, c.cluster.Region))
 
 	DeferCleanup(func(ctx context.Context) {
+
+		err := c.collectLogs(ctx, "bundle", instance.IP)
+		if err != nil {
+			c.logger.Info("issue collecting logs", "error", err)
```
g-gaston, line 484: Does the logger have a `c.logger.Error` method?

g-gaston (approved):

PR #283, opened 2025-01-14
adds cross builds for the e2e bins and ginkgo

jaxesn:
*Issue #, if available:*

*Description of changes:*

Cross build the e2e test binaries (e2e-test is the bin responsible for running setup/clean and e2e.test is the bin which contains the compiled ginkgo tests) and include ginkgo in the artifacts bucket.  This is to be used by the canaries to avoid having to download ginkgo each time and break the arch dependency between the build instance and the canary runner.

Installing ginkgo with go install introduces a bit of weirdness when doing "compilation", but I have commented the code for future me who almost def be confused when i read this in next week!

*Testing (if applicable):*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,9 +105,19 @@ run: build ## Run nodeadm binary.
 e2e-tests-binary: ## Build binary with e2e tests.
 	CGO_ENABLED=0 $(GO) test -ldflags "-s -w -buildid='' -extldflags -static" -c ./test/e2e/suite -o ./_bin/e2e.test -tags "e2e"
 
+.PHONY: build-cross-e2e-tests-binary
+build-cross-e2e-tests-binary:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) test -ldflags "-s -w -buildid='' -extldflags -static" -c ./test/e2e/suite -o ./_bin/amd64/e2e.test -tags "e2e"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -177,6 +187,21 @@ ginkgo: $(GINKGO) ## Download ginkgo.
...
+.PHONY: install-cross-ginkgo
+install-cross-ginkgo: $(LOCALBIN)/amd64/ginkgo $(LOCALBIN)/arm64/ginkgo
+
+# When cross-compiling with go install, you can not override the output directory
+# with GOBIN.  In the case that it is a different arch than the host
+# go install will put the bins in a `goos_goarch` folder under GOPATH/bin
+$(LOCALBIN)/%/ginkgo: GOARCH=$*
+$(LOCALBIN)/%/ginkgo: GOPATH=$(shell $(GO) env GOPATH | cut -d ':' -f 1)
+$(LOCALBIN)/%/ginkgo: CROSS_ARCH_INSTALL_FILE=$(GOPATH)/bin/linux_$(GOARCH)/$(@F)
+$(LOCALBIN)/%/ginkgo: NATIVE_INSTALL_FILE=$(GOPATH)/bin/$(@F)
```
g-gaston, line 199: could we use GOBIN for this or it won't work with cross platform install?

g-gaston (approved):

PR #297, opened 2025-01-17
updates cilium to 1.16.3 for e2e tests

jaxesn:
*Issue #, if available:*

*Description of changes:*

The cilium helm chart tweaks the apparmor config depending on the "server version" of the kubernetes cluster.  For clusters 1.29 and less it uses annotations:

```
  template:
    metadata:
      annotations:
        # Set app AppArmor's profile to "unconfined". The value of this annotation
        # can be modified as long users know which profiles they have available
        # in AppArmor.
        container.apparmor.security.beta.kubernetes.io/cilium-agent: "unconfined"
        container.apparmor.security.beta.kubernetes.io/clean-cilium-state: "unconfined"
        container.apparmor.security.beta.kubernetes.io/mount-cgroup: "unconfined"
        container.apparmor.security.beta.kubernetes.io/apply-sysctl-overwrites: "unconfined"
  ```

For 1.30 and above its on the spec:
```
    spec:
      securityContext:
        appArmorProfile:
          type: Unconfined
```

This PR updates to the new patch of cilium and also adds handling for this difference.  Currently the template uses the 1.29 style, which works for 1.30+, but outputs a deprecation warning.  I think whenever we generated the template file originally we did it a cluster that was 1.29 or less, or an older helm version which defaulted to a lower kube version (im not 100% sure how helm decides what the server version is when generating the template vs installing).

Handling this difference now will avoid future accidental updates, not realizing there is a difference and breaking the < 1.30 versions.

*Testing (if applicable):*

### CODE REVIEW

File: test/e2e/cni/cilium.go [hunk 1]
```diff
@@ -40,7 +44,11 @@ func NewCilium(k8s dynamic.Interface, podCIDR, region string) Cilium {
 }
 
 // Deploy creates or updates the Cilium reosurces.
-func (c Cilium) Deploy(ctx context.Context) error {
+func (c Cilium) Deploy(ctx context.Context, kubernetesVersion string) error {
```
g-gaston, line 47: to be consistent i would pass the kube version in the constructor
the pod cidr is also a cluster attribute and we pass it there

test/e2e/cni/cilium.go
```diff
@@ -12,14 +12,18 @@ import (
...
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/util/retry"
 
 	"github.com/aws/eks-hybrid/test/e2e/constants"
 )
 
-//go:embed testdata/cilium/cilium-template.yaml
-var ciliumTemplate []byte
+//go:embed testdata/cilium/cilium-template-129.yaml
+var ciliumTemplate129 []byte
```
g-gaston, line 23: maybe worth a little comment here pointing to the difference?

g-gaston (approved):

PR #304, opened 2025-01-22
Remove containerd docker repo on uninstall

achen1210:
*Issue #, if available:* 2746

*Description of changes:* 

Removes the docker repo fetched by the package manager upon uninstall when the original containerd-source was 'docker'. 

*Testing (if applicable):*

Tested on Ubuntu for apt and CentOS for yum. 

- Verified that repo is correctly deleted and `apt install containerd.io` and `yum install containerd.io` no longer finds package after nodeadm uninstall.
- Checked that the docker repo is untouched upon uninstall if not originally installed by nodeadm

### DIFF
--- a/cmd/nodeadm/uninstall/uninstall.go
+++ b/cmd/nodeadm/uninstall/uninstall.go
@@ -116,13 +116,6 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		return err
 	}
 
-	// Only removes containerd repos installed by nodeadm (tracked in tracker file)
-	if containerdSource == containerd.ContainerdSourceDocker {
-		if err := packageManager.UninstallPackageManagerDockerRepo(); err != nil {
-			return err
-		}
-	}
-
 	uninstaller := &flows.Uninstaller{
 		Artifacts:      installed.Artifacts,
 		DaemonManager:  daemonManager,
--- a/cmd/nodeadm/uninstall/uninstall.go
+++ b/cmd/nodeadm/uninstall/uninstall.go
@@ -116,6 +116,13 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		return err
 	}
 
+	// Only removes containerd repos installed by nodeadm (tracked in tracker file)
+	if containerdSource == containerd.ContainerdSourceDocker {
+		if err := packageManager.UninstallPackageManagerDockerRepo(); err != nil {
+			return err
+		}
+	}
+
 	uninstaller := &flows.Uninstaller{
 		Artifacts:      installed.Artifacts,
 		DaemonManager:  daemonManager,
--- a/internal/packagemanager/packagemanager.go
+++ b/internal/packagemanager/packagemanager.go
@@ -158,7 +158,7 @@ func (pm *DistroPackageManager) configureAptPackageManagerWithDockerRepo(ctx con
 }
 
 // UninstallPackageManagerDockerRepo uninstalls docker repos installed by package managers when containerd source is docker
-func (pm *DistroPackageManager) uninstallPackageManagerDockerRepo() error {
+func (pm *DistroPackageManager) uninstallDockerRepo() error {
 	removeRepoFile := func(path, pkgType string) error {
 		_, err := os.Stat(path)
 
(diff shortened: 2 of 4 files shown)


cmd/nodeadm/uninstall/uninstall.go
```diff
@@ -116,6 +116,13 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		return err
 	}
 
+	// Only removes containerd repos installed by nodeadm (tracked in tracker file)
+	if containerdSource == containerd.ContainerdSourceDocker {
```
jaxesn, line 120: what is checking the tracking file?

jaxesn (commented):
I dont see any existing uninstall [integration](https://github.com/aws/eks-hybrid/tree/main/test/integration) tests, but maybe this would be a good one to try and add an integration test for?

cmd/nodeadm/uninstall/uninstall.go
achen1210: containerdSource comes from the tracking file: https://github.com/aws/eks-hybrid/pull/304/files/02893e4f163b0378f3d1f533f8f83690805601f5#diff-757a82178c78a849cf0f9b8a27c5af3eedefa82cc5819a851542626f31406553L112

If docker repo was already installed prior nodeadm install, then the tracker file will have containerd source as "" and thus it won't be uninstalled, if it was installed by nodeadm using --containerd-source docker flag then tracker file will have docker as containerd source

vignesh-goutham: This might not be the right place to check and run the uninstall. Lets move this to cleanup section in the internal/flows/uninstall.go file.

cmd/nodeadm/uninstall/uninstall.go
```diff
@@ -116,6 +116,13 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
...
+		if err := packageManager.UninstallPackageManagerDockerRepo(); err != nil {
```
vignesh-goutham, line 121: While today we only install Docker repo, we should have a function that uninstalls any leftovers we mightve installed during installation from package managers perspective. So from the cleanup function on uninstall, we could call a packageManager.Uninstall(), that further calls uninstalling docker repo, if its installed. The check if docker repo is enabled should be done inside the packageManager.`if pm.dockerRepo != "" {...}`

cmd/nodeadm/uninstall/uninstall.go
vignesh-goutham: This might not be the right place to check this. Lets move the check to internal/flows/uninstall.Cleanup()

achen1210: done

cmd/nodeadm/uninstall/uninstall.go
achen1210: completed - good idea

internal/packagemanager/packagemanager.go
```diff
@@ -155,6 +157,43 @@ func (pm *DistroPackageManager) configureAptPackageManagerWithDockerRepo(ctx con
 	return nil
 }
 
+// UninstallPackageManagerDockerRepo uninstalls docker repos installed by package managers when containerd source is docker
+func (pm *DistroPackageManager) uninstallPackageManagerDockerRepo() error {
```
vignesh-goutham, line 161:
```suggestion
func (pm *DistroPackageManager) uninstallDockerRepo() error {
```

internal/packagemanager/packagemanager.go
```diff
@@ -222,6 +261,18 @@ func (pm *DistroPackageManager) runcPackage() artifact.Package {
 	)
 }
 
+// CleanupPackageManagerInstallArtifacts cleans up any artifacts used by package manager during nodeadm install process
+func (pm *DistroPackageManager) CleanupPackageManagerInstallArtifacts() error {
```
vignesh-goutham, line 265:
```suggestion
func (pm *DistroPackageManager) Cleanup() error {
```

vignesh-goutham (commented):
Looks good, tiny nits.

vignesh-goutham (approved):

PR #305, opened 2025-01-23
Rocky support for nodeadm + tests

vignesh-goutham:
*Description of changes:*
Adding rocky support for nodeadm along with few validations and e2e tests

*Testing (if applicable):*
Tested E2E with amd and arm for both Rocky 8 and 9


internal/creds/creds.go
```diff
@@ -42,3 +45,14 @@ func GetCredentialProviderFromInstalledArtifacts(artifacts *tracker.InstalledArt
 	}
 	return "", fmt.Errorf("no credential process found in installed artifacts")
 }
+
+func ValidateCredentialProvider(provider CredentialProvider) error {
```
jaxesn, line 49: lets break this out and merge it as a separate PR, seems like a good change regardless of rocky.

jaxesn: ubuntu 2004 has the same issue

test/e2e/kubernetes/kubernetes.go
```diff
@@ -31,7 +31,7 @@ import (
 const (
 	nodePodWaitTimeout       = 3 * time.Minute
 	nodePodDelayInterval     = 5 * time.Second
-	hybridNodeWaitTimeout    = 10 * time.Minute
+	hybridNodeWaitTimeout    = 12 * time.Minute
```
jaxesn, line 34: rocky slower than rhel?!

go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/aws/eks-hybrid
 
-go 1.22
+go 1.22.0
```
jaxesn, line 3: i suspect u can revret all of this. this happens usually because the cdk node_modules folder exists and go get/mod for some reason sees stuff in there and adds it.

jaxesn (commented):
Im not 100% sure what we want to do with this yet, suggested breaking 1 thing out that we can merge. If there is are other things you think make sense without adding rocky, break those out and we can discuss offline how to proceed with this one.

vignesh-goutham:
Closing this PR for now.

PR #309, opened 2025-01-23
Validate credential provider during install

vignesh-goutham:
*Description of changes:*
Validate the credential provider during install phase. Both Ubuntu 20 and RHEL 8, ship with lower version of GLIBC which doesn't work with iam-roles-anywhere. 

Also has an improvement to the pagination function used with ec2:DescribeImages API in the E2E test.

### DIFF
--- a/internal/creds/validation.go
+++ b/internal/creds/validation.go
@@ -1,11 +1,15 @@
 package creds
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 
 	"github.com/aws/eks-hybrid/internal/api"
 	"github.com/aws/eks-hybrid/internal/iamrolesanywhere"
 	"github.com/aws/eks-hybrid/internal/ssm"
+	"github.com/aws/eks-hybrid/internal/system"
 	"github.com/aws/eks-hybrid/internal/validation"
 )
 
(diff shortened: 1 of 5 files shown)


internal/creds/validation.go
```diff
@@ -23,3 +27,28 @@ func Validations(config aws.Config, node *api.NodeConfig) []validation.Validatio
...
+		}
+
+		// Both RHEL8 and Ubuntu 20 have older version of glibc which iam roles anywhere credential helper doesn't work with
+		// Until we have a fix for that, we will validate and avoid these os version combinations
+		// https://github.com/aws/rolesanywhere-credential-helper/issues/90
+		if (osName == system.RhelOsName && majorOsVersion == "8") || (osName == system.UbuntuOsName && majorOsVersion == "20") {
+			return fmt.Errorf("iam-ra credential provider is not supported on %s %s based operating systems. Please use ssm credential provider", osName, osVersion)
+		}
+	}
+	return nil
```
g-gaston, line 45: nit
```suggestion
	if provider != IamRolesAnywhereCredentialProvider {
	  return nil
	}
	
    majorOsVersion, err := getMajorVersion(osVersion)
    if err != nil {
	    return err
    }
    
    // Both RHEL8 and Ubuntu 20 have older version of glibc which iam roles anywhere credential helper doesn't work with
    // Until we have a fix for that, we will validate and avoid these os version combinations
    // https://github.com/aws/rolesanywhere-credential-helper/issues/90
    if (osName == system.RhelOsName && majorOsVersion == "8") || (osName == system.UbuntuOsName && majorOsVersion == "20") {
	    return fmt.Errorf("iam-ra credential provider is not supported on %s %s based operating systems. Please use ssm credential provider", osName, osVersion)
    }
	
	return nil
```

g-gaston (approved):

PR #311, opened 2025-01-26
adds wait for instance to be registered with ssm

jaxesn:
*Issue #, if available:*

*Description of changes:*

After the setup stack is created, we run an ssm command on the jumpbox to download and verify the ssh private key.  This adds a wait until the instance is registered with SSM to ensure we do not try to run the command, and fail, too soon.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/ssm/command.go
+++ b/test/e2e/ssm/command.go
@@ -8,14 +8,18 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ssm"
+	"github.com/aws/aws-sdk-go-v2/service/ssm/types"
+	smithytime "github.com/aws/smithy-go/time"
 	"github.com/go-logr/logr"
 
 	e2eCommands "github.com/aws/eks-hybrid/test/e2e/commands"
 )
 
 const (
-	commandExecTimeout = 10 * time.Minute
-	commandWaitTimeout = commandExecTimeout + time.Minute
+	commandExecTimeout           = 10 * time.Minute
+	commandWaitTimeout           = commandExecTimeout + time.Minute
+	instanceRegisterTimeout      = 5 * time.Minute
+	instanceRegisterSleepTimeout = 15 * time.Second
 )
 
 // ssm commands run as root user on jumpbox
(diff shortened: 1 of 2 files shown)


test/e2e/ssm/command.go
```diff
@@ -82,3 +88,34 @@ func RunCommand(ctx context.Context, client *ssm.Client, instanceId, command str
 
 	return commandOutput, nil
 }
+
+// there is no built in wait for instance to be registered with ssm
+// WaitForInstance uses DescribeInstanceInformation in a loop to wait for it be registered
+// see: https://github.com/aws/aws-cli/issues/4006
+func WaitForInstance(ctx context.Context, client *ssm.Client, instanceId string, logger logr.Logger) error {
+	waitCtx, cancel := context.WithTimeout(ctx, instanceRegisterTimeout)
```
jaxesn, line 96: Would it be better to pass this in with the timeout? Another option is to do what the aws internal retry code does and keep track of start and end times and keep looping while timeremaining > 0.

test/e2e/ssm/command.go
```diff
@@ -82,3 +88,34 @@ func RunCommand(ctx context.Context, client *ssm.Client, instanceId, command str
 
 	return commandOutput, nil
 }
+
+// there is no built in wait for instance to be registered with ssm
+// WaitForInstance uses DescribeInstanceInformation in a loop to wait for it be registered
+// see: https://github.com/aws/aws-cli/issues/4006
```
g-gaston, line 94:
```suggestion
// WaitForInstance uses DescribeInstanceInformation in a loop to wait for it be registered
// There is no built in wait for instance to be registered with ssm
// see: https://github.com/aws/aws-cli/issues/4006
```

test/e2e/ssm/command.go
g-gaston: I think this is fine for our usecase since we always want to wait the same amount

If we ever want to make the timeout "configurable", we could check if the context passed in has a timeout and if it does, respect it. Else, configure it. This would guarantee that the method doesn't block forever but still allows to override the default timeout.

test/e2e/ssm/command.go
```diff
@@ -82,3 +88,34 @@ func RunCommand(ctx context.Context, client *ssm.Client, instanceId, command str
...
+					Key:    aws.String("InstanceIds"),
+					Values: []string{instanceId},
+				},
+			},
+		})
+		if err != nil {
+			return err
+		}
+		if len(output.InstanceInformationList) > 0 {
+			logger.Info(fmt.Sprintf("Instance %s is registered with SSM", instanceId))
```
g-gaston, line 113: nit: with structured loggers it's usually not recommended to do string interpolation
```suggestion
			logger.Info(fmt.Sprintf("Instance is registered with SSM", "instanceID", instanceId))
```

test/e2e/ssm/command.go
```diff
@@ -8,12 +8,18 @@ import (
...
 	"github.com/go-logr/logr"
 
 	e2eCommands "github.com/aws/eks-hybrid/test/e2e/commands"
 )
 
-const commandTimeout = 10 * time.Minute
+const (
+	commandTimeout               = 10 * time.Minute
+	instanceRegisterTimeout      = 5 * time.Minute
+	instanceRegisterSleepTimeout = 15 * time.Second
```
g-gaston, line 21: is this a random number?
or do instances usually take around 15s to register?

jaxesn: its just random. that is the sleep time if the instance is not register, before it retries the describe.  the overall timeout of 5 mins *should* be plenty, for the jumpbox which we are using this for, its usually up and registered before we ever even get here.

g-gaston (dismissed):

g-gaston (approved):

g-gaston (approved):

PR #317, opened 2025-01-27
breaks e2e test case into mulitiple It blocks to support flakeattempts

jaxesn:
*Issue #, if available:*

*Description of changes:*

This better utilizes the ginkgo test structure by breaking the various parts of the tests into different "nodes".  This also breaks the ec2 creation part of the test out so that it can be retried using FlakeAttempts if for some reason it never goes to running.  This required introducing a new defercleanup strategy to support keeping the instance around between different "tests" but still cleaning it up when appropriate. 

*Testing (if applicable):*

### DIFF
--- a/test/e2e/suite/cleanup.go
+++ b/test/e2e/suite/cleanup.go
@@ -1,0 +1,44 @@
+package suite
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+)
+
+const deferCleanupTimeout = 5 * time.Minute
+
+// TestCleanup is a helper for our use of DeferCleanup in our e2e tests
+// to support retrying EC2 instance creation
+// The cleanup runs after the test itself, just like the default DeferCleanup
+// but it only executes the cleanup if the test was a failure since otherwise
+// we want the instance to stay around for future tests
+// In the BeforeAll TestCleanup.DeferAll is called to register a defer cleanup
+// which runs all registered cleanups from our test methods after all tests
+// are complete
+type TestCleanup struct {
+	cleanups []func(context.Context)
+}
+
+func (t *TestCleanup) Register(f func(context.Context)) {
+	index := len(t.cleanups)
+	t.cleanups = append(t.cleanups, f)
+	DeferCleanup(func(ctx context.Context) {
+		if CurrentSpecReport().Failed() {
+			// remove cleanup if runs due to a test case failure so it does
+			// run again in the deferall
+			t.cleanups = append(t.cleanups[:index], t.cleanups[index+1:]...)
+			f(ctx)
+		}
+	}, NodeTimeout(deferCleanupTimeout))
+}
+
+func (t *TestCleanup) DeferAll() {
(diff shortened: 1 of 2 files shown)


test/e2e/suite/nodeadm_test.go
```diff
@@ -220,109 +217,166 @@ var _ = Describe("Hybrid Nodes", func() {
...
-							Expect(cleanNode.Run(ctx)).To(Succeed(), "node should have been reset successfully")
+								test.logger.Info("Rebooting EC2 Instance.")
+								Expect(nodeadm.RebootInstance(ctx, test.remoteCommandRunner, instance.IP)).NotTo(HaveOccurred(), "EC2 Instance should have rebooted successfully")
+								test.logger.Info("EC2 Instance rebooted successfully.")
+							})
 
-							test.logger.Info("Rebooting EC2 Instance.")
-							Expect(nodeadm.RebootInstance(ctx, test.remoteCommandRunner, instance.IP)).NotTo(HaveOccurred(), "EC2 Instance should have rebooted successfully")
-							test.logger.Info("EC2 Instance rebooted successfully.")
+							It("rejoins as a kubernetes nodes after reboot", func(ctx context.Context) {
```
g-gaston, line 284:
```suggestion
							It("rejoins as a kubernetes node after reboot", func(ctx context.Context) {
```

test/e2e/suite/nodeadm_test.go
```diff
@@ -220,109 +217,166 @@ var _ = Describe("Hybrid Nodes", func() {
...
 
-							test.logger.Info("Waiting for EC2 Instance to be Running...")
-							Expect(ec2.WaitForEC2InstanceRunning(ctx, test.ec2Client, instance.ID)).To(Succeed(), "EC2 Instance should have been reached Running status")
+							It("joins as a kubernetes nodes", func(ctx context.Context) {
+								node, err := kubernetes.WaitForNode(ctx, test.k8sClient, instance.IP, test.logger)
+								Expect(err).NotTo(HaveOccurred(), "Node should have joined the cluster")
+								Expect(node).NotTo(BeNil())
+							})
+
+							It("validates the kubernetes node", FlakeAttempts(2), func(ctx context.Context) {
```
g-gaston, line 262: i would prefer not to repeat this
i understand the appeal but i fear it could hide flakiness in how we setup the node as opposed to flakiness in the validation

test/e2e/suite/cleanup.go
```diff
@@ -0,0 +1,35 @@
...
+	. "github.com/onsi/ginkgo/v2"
+)
+
+const deferCleanupTimeout = 5 * time.Minute
+
+type TestCleanup struct {
+	cleanups []func(context.Context)
+}
+
+func (t *TestCleanup) Register(f func(context.Context)) {
```
g-gaston, line 16: maybe add some doc comments in these two methods to give context on what this does and why is useful?

jaxesn:
Closing for now, the problem with this approach is the ginkgo output gets all intermixed and hard to read.

PR #318, opened 2025-01-27
Add region flag to install

achen1210:
*Issue #, if available:* 2906

*Description of changes:*

Add a --region flag to `nodeadm install`, which will determine which region the SSM installer is downloaded from. 

*Testing (if applicable):*

Tested on ubuntu hybrid node and verified :

- original behavior unchanged when no --region flag used
- region flag changes download location (verified through logging output)
- error is handled appropriately for bad region input by checking download url

Flag will be added to public documentation ahead of nodeadm release.

### DIFF
--- a/cmd/nodeadm/install/install.go
+++ b/cmd/nodeadm/install/install.go
@@ -37,6 +37,7 @@ func NewCommand() cli.Command {
 	fc.AddPositionalValue(&cmd.kubernetesVersion, "KUBERNETES_VERSION", 1, true, "The major[.minor[.patch]] version of Kubernetes to install.")
 	fc.String(&cmd.credentialProvider, "p", "credential-provider", "Credential process to install. Allowed values: [ssm, iam-ra].")
 	fc.String(&cmd.containerdSource, "s", "containerd-source", "Source for containerd artifact. Allowed values: [none, distro, docker].")
+	fc.String(&cmd.region, "r", "region", "AWS region for downloading regional artifacts.")
 	fc.Duration(&cmd.timeout, "t", "timeout", "Maximum install command duration. Input follows duration format. Example: 1h23s")
 	cmd.flaggy = fc
 
@@ -48,15 +49,8 @@ type command struct {
 	kubernetesVersion  string
 	credentialProvider string
 	containerdSource   string
+	region             string
 	timeout            time.Duration
-}
-
-type Config struct {
-	AwsSource          aws.Source
-	ContainerdSource   containerd.SourceName
-	CredentialProvider creds.CredentialProvider
-	Log                *zap.Logger
-	DownloadTimeout    time.Duration
 }
 
 func (c *command) Flaggy() *flaggy.Subcommand {
@@ -113,6 +107,7 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		AwsSource:          awsSource,
 		PackageManager:     packageManager,
 		ContainerdSource:   containerdSource,
+		SsmRegion:          c.region,
 		CredentialProvider: credentialProvider,
 		Logger:             log,
 	}
--- a/internal/flows/install.go
+++ b/internal/flows/install.go
@@ -26,6 +26,7 @@ type Installer struct {
 	ContainerdSource   containerd.SourceName
 	PackageManager     *packagemanager.DistroPackageManager
 	CredentialProvider creds.CredentialProvider
+	SsmRegion          string
 	Tracker            *tracker.Tracker
 	Logger             *zap.Logger
 }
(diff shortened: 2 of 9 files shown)


cmd/nodeadm/install/install.go
```diff
@@ -37,6 +37,7 @@ func NewCommand() cli.Command {
 	fc.AddPositionalValue(&cmd.kubernetesVersion, "KUBERNETES_VERSION", 1, true, "The major[.minor[.patch]] version of Kubernetes to install.")
 	fc.String(&cmd.credentialProvider, "p", "credential-provider", "Credential process to install. Allowed values: [ssm, iam-ra].")
 	fc.String(&cmd.containerdSource, "s", "containerd-source", "Source for containerd artifact. Allowed values: [none, distro, docker].")
+	fc.String(&cmd.region, "r", "region", "AWS region to download install artifacts from.")
```
vignesh-goutham, line 40:
```suggestion
	fc.String(&cmd.region, "r", "region", "AWS region for downloading regional artifacts.")
```

cmd/nodeadm/install/install.go
```diff
@@ -48,13 +49,15 @@ type command struct {
...
 }
 
 type Config struct {
 	AwsSource          aws.Source
 	ContainerdSource   containerd.SourceName
 	CredentialProvider creds.CredentialProvider
+	SsmRegion          string
 	Log                *zap.Logger
 	DownloadTimeout    time.Duration
 }
```
vignesh-goutham, line 63: This struct is deadcode. Can we remove this along with your change?

internal/flows/install.go
```diff
@@ -26,11 +26,20 @@ type Installer struct {
...
 	Tracker            *tracker.Tracker
 	Logger             *zap.Logger
 }
 
+func (i *Installer) Init() {
+	if i.SsmRegion == "" {
+		i.SsmRegion = ssm.DefaultSsmInstallerRegion
+	}
+}
+
```
vignesh-goutham, line 39: Instead of doing it here, how about we set this in the ssm source?

internal/ssm/install.go
```diff
@@ -26,15 +26,17 @@ const (
 // Source serves an SSM installer binary for the target platform.
 type Source interface {
 	GetSSMInstaller(ctx context.Context) (io.ReadCloser, error)
+	GetSSMRegion() string
+	Install(ctx context.Context, t *tracker.Tracker) error
```
vignesh-goutham, line 30: Id like to think the Source interface only deals with anything to do with configuring and getting the actual source of the artifact. That why we didnt initially make Install part of the source interface. I think we can add the `GetSSMRegion()` here, but lets remove Install from the interface.

internal/ssm/source.go
```diff
@@ -17,21 +20,34 @@ const DefaultSsmInstallerRegion = "us-west-2"
 
 // SSMInstaller provides a Source that retrieves the SSM installer from the official
 // release endpoint.
-func NewSSMInstaller(region string) Source {
+func NewSSMInstaller(region string, logger *zap.Logger) Source {
 	return ssmInstallerSource{
```
vignesh-goutham, line 24: Lets check the region here and default if region == ""

internal/ssm/source.go
```diff
@@ -50,6 +66,20 @@ func buildSSMURL(region string) (string, error) {
 	return fmt.Sprintf("https://amazon-ssm-%v.s3.%v.amazonaws.com/latest/%v/ssm-setup-cli", region, region, platform), nil
 }
 
+func validateSSMURL(url string) error {
```
vignesh-goutham, line 69: This is a great check, but I feel by the point this gets to run, it might be late, since some components are already installed and the next step to this is to install SSM. Can we remove this and circle back on checking if the input region is a valid one, as part of install validations?

cmd/nodeadm/install/install.go
achen1210: done

cmd/nodeadm/install/install.go
achen1210: done

internal/flows/install.go
achen1210: done

internal/ssm/install.go
achen1210: done

internal/ssm/source.go
achen1210: done

internal/ssm/source.go
achen1210: done

test/integration/cases/install-with-region/run.sh
```diff
@@ -0,0 +1,52 @@
...
+
+source /helpers.sh
+
+mock::aws
+wait::dbus-ready
+
+# remove previously installed containerd to test installation via nodeadm
+dnf remove -y containerd
+
+declare SUPPORTED_VERSIONS=(1.26 1.27 1.28 1.29 1.30)
```
g-gaston, line 15: 1.31 1.32? although does it matter for this case?

PR #282, opened 2025-01-13
E2E test verify pod identity agent is installed

junpengdev:
*Issue #, if available:*
https://github.com/aws/eks-anywhere-internal/issues/2799

*Description of changes:*
This is part of work to add e2e tests for EKS-Hybrid node. This PR verifies that Pod Identity add-on are successfully created in hybrid nodes.

*Testing (if applicable):*
Run e2e testing.

Sample output:
```
2025-01-13T11:50:23.313-0800	INFO	Successfully validated log output	{"pod": "nginx-mi-03a9903089ef4f22c"}
2025-01-13T11:50:23.314-0800	INFO	Deleting test pod	{"pod": "nginx-mi-03a9903089ef4f22c"}
2025-01-13T11:50:28.348-0800	INFO	Pod deleted successfully	{"pod": "nginx-mi-03a9903089ef4f22c"}
2025-01-13T11:50:28.348-0800	INFO	Testing Pod Identity add-on functionality
2025-01-13T11:50:28.348-0800	INFO	Verify pod identity add-on is installed
2025-01-13T11:50:28.348-0800	INFO	Describe cluster add-on	{"ClusterAddon": "eks-pod-identity-agent"}
2025-01-13T11:50:28.630-0800	INFO	Check if daemon set exists	{"daemonSet": "eks-pod-identity-agent-hybrid"}
2025-01-13T11:50:28.643-0800	INFO	Resetting hybrid node...
```

*Documentation added/planned (if applicable):*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice

PR #329, opened 2025-01-29
Set up pod identity association in E2E tests

junpengdev:
*Issue #, if available:*
This is part 2 of setting up E2E tests for pod identity. [Part 1 PR](https://github.com/aws/eks-hybrid/pull/282)

*Description of changes:*
This PR covers pod identity association setup. 

*(Since this PR already contains enough changes, I will create another PR to actually deploy pods and verify EKS Pod Identity volume is correctly attached.)*

I followed this [public document](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-association.html#_create_a_pod_identity_association_cli) to set up pod identity association.

In summary, the following is required in order to set up pod identity association.
1. Create IAM role with trust relationship for Pod Identity role. The role should also be able to access some AWS resource.
2. Create a service account dedicated for Pod Identity
3. Associate the IAM role to the service account

---
The creation of service account and pod identity association can only be done after the EKS cluster is provisioned successfully. We can put this into either 
1) the creation of Pod Identity add-on, or
2) during the add-on e2e tests.

Here I choose to put it into creation of pod identity add-on. The reason for it is we don't need to check or set up this association in every `OS/auth provider` combination. It should be a one-time setup for the whole EKS cluster.

---

The creation of IAM role of pod identity can be put into
1) `nodeadm-stack.ts` to provision one IAM role for all `kube/CNI` combination
2) `setup-cfn.yaml` to provision one IAM role for all `OS/Auth provider` under one `kube/CNI` combination
3) during e2e test to check or provision one IAM role for each `OS/Auth provider` 

Option 3 is not applicable for two reasons.
- pod identity association is now done during creation of the add-on, which requires IAM ARN to complete the process
- it's very inefficient to check/provision the IAM role for each `OS/Auth provider`

Option 1 and 2 are both working but I choose option 2 because it makes more sense as the role is tied to one cluster, which the lifecycle is managed by CloudFormation.

---
*Testing (if applicable):*
1. Run e2e test locally
5. Deploy the change to personal dev stack in CodePipeline

### DIFF
--- a/test/e2e/addon/podidentityaddon.go
+++ b/test/e2e/addon/podidentityaddon.go
@@ -1,0 +1,53 @@
+package addon
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/eks"
+	"github.com/aws/aws-sdk-go-v2/service/eks/types"
+	"github.com/aws/aws-sdk-go-v2/service/iam"
+	"github.com/go-logr/logr"
+	clientgo "k8s.io/client-go/kubernetes"
+
+	"github.com/aws/eks-hybrid/test/e2e/errors"
+	"github.com/aws/eks-hybrid/test/e2e/kubernetes"
+)
+
+type PodIdentityAddon struct {
+	Addon
+	Kubernetes         *clientgo.Clientset
+	IAMClient          *iam.Client
+	PodIdentityRoleArn string
+}
+
+const (
+	podIdentityServiceAccount = "pod-identity-sa"
+	namespace                 = "default"
+)
+
+func (p PodIdentityAddon) Create(ctx context.Context, client *eks.Client, logger logr.Logger) error {
+	if err := p.Addon.Create(ctx, client, logger); err != nil {
+		return err
+	}
+
+	// Provision PodIdentity addon related resources
+	// Create service account in kubernetes
+	if err := kubernetes.NewServiceAccount(ctx, logger, p.Kubernetes, namespace, podIdentityServiceAccount); err != nil {
+		return err
+	}
+
+	createPodIdentityAssociationInput := &eks.CreatePodIdentityAssociationInput{
+		ClusterName:    &p.Cluster,
+		Namespace:      aws.String(namespace),
+		RoleArn:        &p.PodIdentityRoleArn,
+		ServiceAccount: aws.String(podIdentityServiceAccount),
+	}
+
+	_, err := client.CreatePodIdentityAssociation(ctx, createPodIdentityAssociationInput)
+	if err != nil && !errors.IsType(err, &types.ResourceInUseException{}) {
+		return err
+	}
+
+	return nil
+}
--- a/test/e2e/addon/podidentityaddon.go
+++ b/test/e2e/addon/podidentityaddon.go
@@ -6,7 +6,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/eks"
 	"github.com/aws/aws-sdk-go-v2/service/eks/types"
-	"github.com/aws/aws-sdk-go-v2/service/iam"
 	"github.com/go-logr/logr"
 	clientgo "k8s.io/client-go/kubernetes"
 
@@ -16,9 +15,7 @@ import (
 
 type PodIdentityAddon struct {
 	Addon
-	kubernetes *clientgo.Clientset
-	iamClient  *iam.Client
-	roleArn    string
+	roleArn string
 }
 
 const (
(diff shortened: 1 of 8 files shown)


test/e2e/addon/podidentityverify.go
```diff
@@ -27,19 +28,27 @@ type VerifyPodIdentityAddon struct {
 func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
 	v.Logger.Info("Verify pod identity add-on is installed")
 
-	podIdentityAddon := Addon{
-		Name:    podIdentityAgent,
-		Cluster: v.Cluster,
+	podIdentityAddon := PodIdentityAddon{
```
g-gaston, line 31: can this just be a `Addon` instead?
it seems like we don't need the pod identity struct since we only want to wait until it's active?

test/e2e/addon/podidentityverify.go
```diff
@@ -27,19 +28,27 @@ type VerifyPodIdentityAddon struct {
...
+			Cluster: v.Cluster,
+		},
 	}
 
 	timeoutCtx, cancel := context.WithTimeout(ctx, getAddonTimeout)
 	defer cancel()
 
-	if err := podIdentityAddon.WaitUtilActive(timeoutCtx, v.EKSClient, v.Logger); err != nil {
+	var err error
+	if err = podIdentityAddon.WaitUtilActive(timeoutCtx, v.EKSClient, v.Logger); err != nil {
```
g-gaston, line 42: curious why this change?

test/e2e/cluster/create.go
```diff
@@ -110,6 +103,27 @@ func (c *Create) Run(ctx context.Context, test TestResources) error {
 		return fmt.Errorf("loading kubeconfig: %w", err)
 	}
 
+	k8sClient, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return fmt.Errorf("creating kubernetes client: %w", err)
+	}
+
+	podIdentityAddon := addon.PodIdentityAddon{
```
g-gaston, line 111: could we have a constructor for this?

I'm not against creating this "service" objects with a struct directly, but in this case it seems like to do so we have to reveal a lot of the internal details: the fact that wraps an Addon, the name of the add-on, the configuration... It appears that those details could be hidden in a constructor so they remain internal to the addon package.

test/e2e/cluster/create.go
```diff
@@ -110,6 +103,27 @@ func (c *Create) Run(ctx context.Context, test TestResources) error {
...
+			Name:          podIdentityAddonName,
+			Configuration: "{\"daemonsets\":{\"hybrid\":{\"create\": true}}}",
+		},
+		Kubernetes:         k8sClient,
+		IAMClient:          c.iam,
+		PodIdentityRoleArn: stackOut.podIdentityRoleArn,
+	}
+
+	err = podIdentityAddon.Create(ctx, c.eks, c.logger)
+	if err != nil && !errors.IsType(err, &types.ResourceInUseException{}) {
```
g-gaston, line 123: aren't you handling that error now in the `Create` method itself? it does seem like it's a good thing to hide if the intent is to make the create operation idempotent

test/e2e/kubernetes/kubernetes.go
```diff
@@ -544,3 +544,27 @@ func GetDaemonSet(ctx context.Context, logger logr.Logger, k8s *kubernetes.Clien
 
 	return foundDaemonSet, nil
 }
+
+func NewServiceAccount(ctx context.Context, logger logr.Logger, k8s *kubernetes.Clientset, namespace, name string) error {
+	if _, err := k8s.CoreV1().ServiceAccounts(namespace).Get(ctx, name, metav1.GetOptions{}); err == nil {
+		logger.Info("Service account has already been created", "namespace", namespace, name)
```
g-gaston, line 550: super nit
```suggestion
		logger.Info("Service account already exists", "namespace", namespace, "name", name)
```

test/e2e/addon/podidentityverify.go
junpengdev: This is to ensure the add-on is in Active state.

test/e2e/addon/podidentityaddon.go
```diff
@@ -0,0 +1,66 @@
...
+			Name:          name,
+			Configuration: "{\"daemonsets\":{\"hybrid\":{\"create\": true}}}",
+		},
+		kubernetes: k8sClient,
+		iamClient:  iamClient,
+		roleArn:    roleArn,
+	}
+}
+
+func (p PodIdentityAddon) Create(ctx context.Context, client *eks.Client, logger logr.Logger) error {
```
g-gaston, line 42: why is the eks client passed an argument here but the iam client and k8s client passed to the constructor?

i would vote to pass them all to this method (if you need them) and make the struct "data only"

test/e2e/addon/podidentityverify.go
g-gaston: sorry, i wasn't very clear. I'm asking why we went from the one liner to defining the error separatly

test/e2e/addon/podidentityaddon.go
```diff
@@ -0,0 +1,61 @@
...
+	Addon
+	roleArn string
+}
+
+const (
+	podIdentityServiceAccount = "pod-identity-sa"
+	namespace                 = "default"
+)
+
+func NewPodIdentityAddon(cluster, name, roleArn string) PodIdentityAddon {
```
g-gaston, line 26: Isn't the addon always the same? If so, why does the caller need to provide? This constructor is already exclusive for the pod identity addon, so i think it encapsulate de knowlege of the addon name and hide that detail from the caller

Wdyt?

g-gaston (approved):

test/e2e/kubernetes/kubernetes.go
```diff
@@ -544,3 +544,27 @@ func GetDaemonSet(ctx context.Context, logger logr.Logger, k8s *kubernetes.Clien
...
+			APIVersion: "v1",
+			Kind:       "ServiceAccount",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+		},
+	}
+
+	if _, err := k8s.CoreV1().ServiceAccounts(namespace).Create(ctx, serviceAccount, metav1.CreateOptions{}); err != nil {
```
jaxesn, line 565: I think we probably should probably retry this

junpengdev: Let's keep it this way for now. I'd like to understand what errors we may have and what errors are retryable.

test/e2e/addon/podidentityaddon.go
junpengdev: good suggestion. I will add it to my part3 PR.

PR #332, opened 2025-01-29
Add region flag to install

achen1210:
*Issue #, if available:* 2906

*Description of changes:*

Add a --region flag to nodeadm install, which will determine which region the SSM installer is downloaded from.

*Testing (if applicable):*

Tested on ubuntu hybrid node and verified :

original behavior unchanged when no --region flag used
region flag changes download location (verified through logging output)
error is handled appropriately for bad region input by checking download url

Flag will be added to public documentation ahead of nodeadm release.


test/integration/cases/install-with-region/run.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# remove previously installed containerd to test installation via nodeadm
+dnf remove -y containerd
+
+declare SUPPORTED_VERSIONS=(1.26 1.27 1.28 1.29 1.30)
+
+for VERSION in ${SUPPORTED_VERSIONS}
+do
+
+    nodeadm install $VERSION --credential-provider ssm --region us-east-1
```
jaxesn, line 20: is it worth asserting somehow, maybe on the output, that the region passed in was actually used?

achen1210: done

cmd/nodeadm/install/install.go
```diff
@@ -38,6 +38,7 @@ func NewCommand() cli.Command {
 	fc.AddPositionalValue(&cmd.kubernetesVersion, "KUBERNETES_VERSION", 1, true, "The major[.minor[.patch]] version of Kubernetes to install.")
 	fc.String(&cmd.credentialProvider, "p", "credential-provider", "Credential process to install. Allowed values: [ssm, iam-ra].")
 	fc.String(&cmd.containerdSource, "s", "containerd-source", "Source for containerd artifact. Allowed values: [none, distro, docker].")
+	fc.String(&cmd.region, "r", "region", "AWS region for downloading regional artifacts.")
```
g-gaston, line 41: I would vote to set here a default so it shows up in the command docs

internal/ssm/source.go
```diff
@@ -31,16 +35,22 @@ type ssmInstallerSource struct {
...
 	}
-	s.logger.Info("Downloading SSM installer", zap.String("url", endpoint))
+
+	s.logger.Info("Downloading SSM installer", zap.String("region", s.region), zap.String("url", endpoint))
+
 	obj, err := util.GetHttpFileReader(ctx, endpoint)
 	if err != nil {
-		obj.Close()
-		return nil, err
+		return nil, fmt.Errorf("unable to download SSM installer from region %s: "+
```
g-gaston, line 52:
```suggestion
		return nil, fmt.Errorf("unable to download SSM installer from region %s, "+
```

test/integration/cases/install-with-region/run.sh
```diff
@@ -0,0 +1,53 @@
...
+
+mock::aws
+wait::dbus-ready
+
+# remove previously installed containerd to test installation via nodeadm
+dnf remove -y containerd
+
+declare SUPPORTED_VERSIONS=(1.26 1.27 1.28 1.29 1.30)
+
+for VERSION in ${SUPPORTED_VERSIONS}
```
g-gaston, line 17: if you want to run this in a loop, don't you have to run uninstall at the end of the loop?

tbh for this test idk if the k8s versions matter. the ssm installation is the same for all of them

test/integration/helpers.sh
```diff
@@ -153,6 +153,46 @@ function assert::file-permission-matches() {
...
+    local arch=$(uname -m)
+
+    # Convert architecture names to match SSM's format
+    case "$arch" in
+        "x86_64")  arch="linux_amd64" ;;
+        "aarch64") arch="linux_arm64" ;;
+        *) echo "Unsupported architecture: $arch" >&2; return 1 ;;
+    esac
+
+    local expected_url="https://amazon-ssm-${region}.s3.${region}.amazonaws.com/latest/${arch}/ssm-setup-cli"
```
g-gaston, line 188: i don't love relying on the logs to test something

I think the integration test is great, but i think we should just test that it actually downloads the installer and that the agent gets installed by the installer by just looking at the files on disk

If you want to verify what. is the exact url that the code generates for a given input, let's write some unit tests for it.

internal/ssm/install.go
```diff
@@ -116,12 +117,13 @@ func uninstallPreRegisterComponents(ctx context.Context, pkgSource PkgSource) er
...
 }
 
-func runInstallWithRetries(ctx context.Context) error {
+func runInstallWithRetries(ctx context.Context, region string) error {
 	// Sometimes install fails due to conflicts with other processes
 	// updating packages, specially when automating at machine startup.
 	// We assume errors are transient and just retry for a bit.
 	installCmdBuilder := func(ctx context.Context) *exec.Cmd {
-		return exec.CommandContext(ctx, installerPath, "-install", "-region", DefaultSsmInstallerRegion)
+		return exec.CommandContext(ctx, installerPath, "-install", "-region", region)
```
g-gaston, line 125: how difficult would it be to write a test that checks this?

internal/flows/install.go
```diff
@@ -78,9 +79,7 @@ func (i *Installer) installCredentialProcess(ctx context.Context) error {
 			return err
 		}
 	case creds.SsmCredentialProvider:
-		ssmInstaller := ssm.NewSSMInstaller(i.Logger, ssm.DefaultSsmInstallerRegion)
-
-		i.Logger.Info("Installing SSM agent installer...")
```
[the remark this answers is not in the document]
achen1210: I think it felt redundant since we already had a message about downloading the SSM installer https://github.com/aws/eks-hybrid/blob/main/internal/ssm/source.go#L109 but we can leave it in too

internal/ssm/source.go
```diff
@@ -19,7 +19,11 @@ const DefaultSsmInstallerRegion = "us-west-2"
 
 // SSMInstaller provides a Source that retrieves the SSM installer from the official
 // release endpoint.
-func NewSSMInstaller(logger *zap.Logger, region string) Source {
+func NewSSMInstaller(region string, logger *zap.Logger) Source {
```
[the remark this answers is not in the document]
achen1210: we can leave it as is, I originally changed it since I felt like the region is critical to the function so it should come first and the logger is more auxiliary

test/integration/cases/install-with-region/run.sh
achen1210: yeah it makes more sense to run it for only 1 version

test/integration/helpers.sh
achen1210: yeah makes sense, will add

internal/ssm/install.go
```diff
@@ -26,6 +26,7 @@ const (
 // Source serves an SSM installer binary for the target platform.
 type Source interface {
 	GetSSMInstaller(ctx context.Context) (io.ReadCloser, error)
+	GetSSMRegion() string
```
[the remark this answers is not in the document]
achen1210: yeah I see makes sense I'll change it to that

internal/ssm/install.go
achen1210: will include as part of test comment below

vignesh-goutham (approved):

internal/ssm/install.go
```diff
@@ -99,6 +100,9 @@ func downloadFileTo(ctx context.Context, source Source, installerPath string) er
 		return fmt.Errorf("installing ssm-setup-cli: %w", err)
 	}
 
+	if err = runInstallWithRetries(ctx, installerPath, region); err != nil {
```
g-gaston, line 103: are you running install twice? once from downloadFileTo and once from top level Install?

internal/ssm/source.go
```diff
@@ -78,6 +78,9 @@ func WithPublicKey(key string) SSMInstallerOption {
 // SSMInstaller provides a Source that retrieves the SSM installer from the official
 // release endpoint.
 func NewSSMInstaller(logger *zap.Logger, region string, opts ...SSMInstallerOption) Source {
+	if region == "" {
```
g-gaston, line 81: now that we have a default in the cmd level, can we remove this? just make region by mandatory
if you want, return error/panic if region is empty

this would reduce the chances of us breaking the flow and endup passing an empty string here and silently using the default region

test/integration/cases/install-with-region/run.sh
```diff
@@ -0,0 +1,55 @@
...
+set -o pipefail
+
+source /helpers.sh
+
+VERSION=1.31
+
+mock::aws
+wait::dbus-ready
+# block access to us-west-2 ssm install url
+echo "0.0.0.0 amazon-ssm-us-west-2.s3.us-west-2.amazonaws.com" >> /etc/hosts
```
g-gaston, line 14: you could verify that this works by trying to execute it without passing a region and assert it fails

internal/ssm/install.go
achen1210: will remove in a future pr

internal/ssm/source.go
achen1210: will add to future pr

PR #342, opened 2025-02-05
Upgrade improvements - Only upgrade artifacts and services if there is an upgrade avilable

vignesh-goutham:
*Description of changes:*
Upgrade today runs an uninstall followed by install + init. While this does achieve upgrade this run operations that might not be required. With this change upgrade operation will check if a particular artifact is due for an upgrade and then only upgrade the artifact. 

Callouts with this change
- cni-plugins ship as tarball so its not possible to check if there is really an upgrade for the plugins. We always upgrade the plugins with this change.
- Similar situation with the SSM installer cli, which doesnt publish its checksums in order for us to validate it. (Future task to add checksums to manifest that will enable us to do this. The installer is also regional)
- For packages installed from OS's package manager, we run the yum/apt upgrade command which will only upgrade if there is an upgrade available.

Also added a package manager metadata refresh command. I'll swap yum/apt updates with this command in other places in a following change.
 
*Testing (if applicable):*
Tested with E2E tests.

### DIFF
--- a/internal/artifact/install.go
+++ b/internal/artifact/install.go
@@ -1,16 +1,12 @@
 package artifact
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"io/fs"
 	"os"
 	"os/exec"
 	"path"
-	"time"
-
-	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
 // DefaultDirPerms are the permissions assigned to a directory when an Install* func is called
@@ -53,16 +49,4 @@ func InstallTarGz(dst, src string) error {
 		return err
 	}
 	return nil
-}
-
-// InstallPackageWithRetries installs a package and retries errors until the context is
-// cancelled. The backoff duration is the time to wait between retries.
-func InstallPackageWithRetries(ctx context.Context, pkgSource Package, backoff time.Duration) error {
-	return cmd.Retry(ctx, pkgSource.InstallCmd, backoff)
-}
-
-// UninstallPackageWithRetries uninstalls a package and retries errors until the context is
-// cancelled. The backoff duration is the time to wait between retries.
-func UninstallPackageWithRetries(ctx context.Context, pkgSource Package, backoff time.Duration) error {
-	return cmd.Retry(ctx, pkgSource.UninstallCmd, backoff)
 }
--- a/internal/artifact/upgrade_test.go
+++ b/internal/artifact/upgrade_test.go
@@ -1,0 +1,120 @@
+package artifact
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"os"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"go.uber.org/zap"
+)
+
+func TestUpgradeAvailable(t *testing.T) {
+	dummyFilePath := "testdata/dummyfile"
+	dummyFh, err := os.Open(dummyFilePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	fileChecksum := []byte("b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9  internal/artifact/testdata/dummyfile")
+	wrongChecksum := []byte("b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7acabcdcde9 randomfile/path")
+
+	testcases := []struct {
+		name           string
+		filePath       string
+		sourceChecksum []byte
+		checksumMatch  bool
+		wantErr        error
+	}{
+		{
+			name:           "Upgrade available",
+			filePath:       dummyFilePath,
+			sourceChecksum: wrongChecksum,
+			checksumMatch:  false,
+			wantErr:        nil,
+		},
+		{
+			name:           "Upgrade not available",
+			filePath:       dummyFilePath,
+			sourceChecksum: fileChecksum,
+			checksumMatch:  true,
+			wantErr:        nil,
+		},
+		{
+			name:           "File not installed",
+			filePath:       "wrong/path",
+			sourceChecksum: wrongChecksum,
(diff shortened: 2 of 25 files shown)


g-gaston (commented):
Are there any scenarios we might want to add integration tests for?
Or extra assertions to the ones we have?

internal/artifact/install.go
```diff
@@ -66,3 +66,9 @@ func InstallPackageWithRetries(ctx context.Context, pkgSource Package, backoff t
 func UninstallPackageWithRetries(ctx context.Context, pkgSource Package, backoff time.Duration) error {
 	return cmd.Retry(ctx, pkgSource.UninstallCmd, backoff)
 }
+
+// UpgradePackageWithRetries upgrades a package (if available) and retries errors until the context
+// is cancelled. The backoff duration is the time to wait between retries.
+func UpgradePackageWithRetries(ctx context.Context, pkgSource Package, backoff time.Duration) error {
```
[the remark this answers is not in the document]
vignesh-goutham: Yea provides some abstraction but I dont see any good value vs calling cmd.Retry directly.

internal/flows/upgrade.go
```diff
@@ -2,22 +2,142 @@ package flows
...
+		return err
+	}
+
+	if err := u.NodeProvider.ConfigureAws(ctx); err != nil {
+		return err
+	}
+	if err := u.NodeProvider.Enrich(ctx); err != nil {
+		return err
+	}
+	if err := initDaemons(ctx, u.NodeProvider, u.SkipPhases, u.Logger); err != nil {
```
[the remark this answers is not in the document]
vignesh-goutham: Thats right, for now, we dont have know if the user added additional containerd/kubelet configuration, and if they did services need to be restarted for those additional configuration to take effect. I going to work on a way to identify or rather diff between current vs generated config to see if respective services need to be restarted.

internal/flows/upgrade.go
```diff
@@ -2,22 +2,131 @@ package flows
...
+	}
+
+	if err := u.upgradeCredentialProvider(ctx); err != nil {
+		return err
+	}
+
+	if err := u.upgradeEksArtifacts(ctx); err != nil {
+		return err
+	}
+
```
g-gaston, line 50: do we need to call `ValidateConfig` here? or even before

internal/iamauthenticator/install.go
```diff
@@ -43,3 +51,13 @@ func Install(ctx context.Context, tracker *tracker.Tracker, iamAuthSrc IAMAuthen
 func Uninstall() error {
 	return os.RemoveAll(IAMAuthenticatorBinPath)
 }
+
+func Upgrade(ctx context.Context, src IAMAuthenticatorSource, log *zap.Logger) error {
```
g-gaston, line 55: same here about those helper to download files, maybe we can use those to test these

test/integration/helpers.sh
```diff
@@ -86,6 +86,27 @@ function assert::file-not-contains() {
...
 }
 
+function assert::checksum-match() {
+    if [ "$#" -ne 1 ]; then
+      echo "Usage: assert::checksum-match FILE"
+      exit 1
+    fi
+    local INPUT_PATH=$1
+    if sha256sum -c $INPUT_PATH.sha256; then
+      echo "Checksums should not match, bu they do!"
```
g-gaston, line 96: should this method be named `assert::checksum-not-match()` then?

test/integration/cases/upgrade-with-iam/run.sh
```diff
@@ -61,6 +61,9 @@ validate-file /etc/kubernetes/pki/ca.crt 644 expected-ca-crt
 # Order of items in this file is random, skip checking content of /etc/eks/kubelet/environment
 validate-file /etc/eks/kubelet/environment 644
 
+# Since we are upgrading kubernetes version primarily also check if the checksums of artifacts changed
+generate::checksum-file /usr/bin/kubelet
+generate::checksum-file /usr/local/bin/kubectl
```
g-gaston, line 66: how can we introduce some assertions that help us verify that this not doing anymore what we intended to fix with this work?

Like for example, verify not deleting stuff from the cni folder (which i think we already did before but hacky) or verify that we don't removing packages like ssm/containerd...

internal/ssm/install.go
```diff
@@ -24,13 +24,15 @@ import (
...
 	configRoot           = "/etc/amazon"
+	artifactName         = "ssm"
 )
 
 // Source serves an SSM installer binary for the target platform.
 type Source interface {
 	GetSSMInstaller(ctx context.Context) (io.ReadCloser, error)
 	GetSSMInstallerSignature(ctx context.Context) (io.ReadCloser, error)
 	PublicKey() string
+	Region() string
```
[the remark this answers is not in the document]
vignesh-goutham: I see what youre saying. I agree.

internal/flows/upgrade.go
vignesh-goutham: We run it much before since we need to fast fail in the cmd package itself. I'm re-thinking about the validations we have in the cmd package too, I think that needs to be organized well. But thats for another PR.

internal/artifact/upgrade_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+			// Check upgraded written data
+			fileData, err := os.ReadFile(dummyFilePath)
+			if err != nil {
+				t.Fatal(err)
+			}
+			g.Expect(fileData).To(Equal([]byte(tt.upgradedData)))
+
+			// Reset file
+			err = os.WriteFile(dummyFilePath, []byte(installedFileData), 0o755)
```
g-gaston, line 112: does this mean we modify the actual golden file? if so, i would vote to create a different tmp file per testcase. you can just copy the contents of the golden file if that's what we need in there

g-gaston (approved):

PR #345, opened 2025-02-07
Create test pod to verify it can access AWS resources with pod identity add-on

junpengdev:
*Issue #, if available:*
This is part 3 of setting up E2E tests for pod identity. 
[part 1](https://github.com/aws/eks-hybrid/pull/282)
[part 2](https://github.com/aws/eks-hybrid/pull/329)

*Description of changes:*
This PR would create a test pod with a service account, which is associated with pod identity. It then downloads test file from S3 bucket within the test pod. 

[Ref doc](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-how-it-works.html)

*Testing (if applicable):*

### DIFF
--- a/example/hybrid-ira-cfn.yaml
+++ b/example/hybrid-ira-cfn.yaml
@@ -48,7 +48,8 @@ Resources:
     Properties:
       RoleName: !Ref RoleName
       ManagedPolicyArns:
-        - 'arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly'
+        - arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly
+        - arn:aws:iam::aws:policy/AmazonEKSWorkerNodeMinimalPolicy
       Policies:
         - PolicyName: EKSDescribeCluster
           PolicyDocument:
--- a/example/hybrid-ira-cfn.yaml
+++ b/example/hybrid-ira-cfn.yaml
@@ -49,7 +49,6 @@ Resources:
       RoleName: !Ref RoleName
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly
-        - arn:aws:iam::aws:policy/AmazonEKSWorkerNodeMinimalPolicy
       Policies:
         - PolicyName: EKSDescribeCluster
           PolicyDocument:
--- a/test/e2e/addon/podidentityaddon.go
+++ b/test/e2e/addon/podidentityaddon.go
@@ -2,6 +2,7 @@ package addon
 
 import (
 	"context"
+	"errors"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -11,7 +12,6 @@ import (
 	"github.com/go-logr/logr"
 	clientgo "k8s.io/client-go/kubernetes"
 
-	"github.com/aws/eks-hybrid/test/e2e/errors"
 	"github.com/aws/eks-hybrid/test/e2e/kubernetes"
 )
 
@@ -58,14 +58,14 @@ func (p PodIdentityAddon) Create(ctx context.Context, logger logr.Logger, eksCli
 	}
 
 	_, err := eksClient.CreatePodIdentityAssociation(ctx, createPodIdentityAssociationInput)
-	if err != nil && !errors.IsType(err, &types.ResourceInUseException{}) {
-		return err
+	if err == nil || errors.Is(err, &types.ResourceInUseException{}) {
+		return nil
 	}
 
-	return nil
+	return err
 }
 
-func (p PodIdentityAddon) Upload(ctx context.Context, logger logr.Logger, client *s3.Client, bucket string) error {
+func (p PodIdentityAddon) UploadFileForVerification(ctx context.Context, logger logr.Logger, client *s3.Client, bucket string) error {
 	if _, err := client.PutObject(ctx, &s3.PutObjectInput{
 		Bucket: aws.String(bucket),
 		Key:    aws.String(bucketObjectKey),
--- a/test/e2e/addon/podidentityaddon.go
+++ b/test/e2e/addon/podidentityaddon.go
@@ -21,13 +21,14 @@ type PodIdentityAddon struct {
 const (
 	podIdentityServiceAccount = "pod-identity-sa"
 	namespace                 = "default"
+	podIdentityAgent          = "eks-pod-identity-agent"
 )
 
-func NewPodIdentityAddon(cluster, name, roleArn string) PodIdentityAddon {
+func NewPodIdentityAddon(cluster, roleArn string) PodIdentityAddon {
 	return PodIdentityAddon{
 		Addon: Addon{
 			Cluster:       cluster,
-			Name:          name,
+			Name:          podIdentityAgent,
 			Configuration: "{\"daemonsets\":{\"hybrid\":{\"create\": true}}}",
 		},
 		roleArn: roleArn,
--- a/test/e2e/addon/podidentityaddon.go
+++ b/test/e2e/addon/podidentityaddon.go
@@ -2,10 +2,12 @@ package addon
 
 import (
 	"context"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/eks"
 	"github.com/aws/aws-sdk-go-v2/service/eks/types"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/go-logr/logr"
 	clientgo "k8s.io/client-go/kubernetes"
 
@@ -22,6 +24,8 @@ const (
 	podIdentityServiceAccount = "pod-identity-sa"
 	namespace                 = "default"
 	podIdentityAgent          = "eks-pod-identity-agent"
+	bucketObjectKey           = "test"
+	bucketObjectContent       = "RANDOM-WORD"
 )
 
 func NewPodIdentityAddon(cluster, roleArn string) PodIdentityAddon {
@@ -55,6 +59,18 @@ func (p PodIdentityAddon) Create(ctx context.Context, logger logr.Logger, eksCli
 
 	_, err := eksClient.CreatePodIdentityAssociation(ctx, createPodIdentityAssociationInput)
 	if err != nil && !errors.IsType(err, &types.ResourceInUseException{}) {
+		return err
+	}
+
+	return nil
+}
+
+func (p PodIdentityAddon) Upload(ctx context.Context, logger logr.Logger, client *s3.Client, bucket string) error {
+	if _, err := client.PutObject(ctx, &s3.PutObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(bucketObjectKey),
+		Body:   strings.NewReader(bucketObjectContent),
+	}); err != nil {
 		return err
 	}
 
--- a/test/e2e/addon/podidentityverify.go
+++ b/test/e2e/addon/podidentityverify.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/eks"
@@ -25,6 +26,8 @@ const (
 	policyName           = "pod-identity-association-role-policy"
 	PodIdentityS3Bucket  = "PodIdentityS3Bucket"
 )
+
+var PodIdentityS3BucketPrefix = strings.ToLower(PodIdentityS3Bucket)
 
 type VerifyPodIdentityAddon struct {
 	Cluster             string
--- a/test/e2e/addon/podidentityverify.go
+++ b/test/e2e/addon/podidentityverify.go
@@ -3,10 +3,7 @@ package addon
 import (
 	"context"
 	_ "embed"
-	"encoding/json"
 	"fmt"
-	"net/url"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -27,6 +24,7 @@ const (
 	podIdentityDaemonSet = "eks-pod-identity-agent-hybrid"
 	podIdentityToken     = "eks-pod-identity-token"
 	policyName           = "pod-identity-association-role-policy"
+	PodIdentityS3Bucket  = "PodIdentityS3Bucket"
 )
 
 type VerifyPodIdentityAddon struct {
@@ -71,7 +69,7 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
 		return err
 	}
 
-	bucket, err := getBucketNameFromPodIdentityAssociation(ctx, v.Cluster, v.EKSClient, v.IAMClient)
+	bucket, err := getPodIdentityS3Bucket(ctx, v.Cluster, v.S3Client)
 	if err != nil {
 		return err
 	}
@@ -83,7 +81,7 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
 		return err
 	}
 
-	podName := "awscli"
+	podName := fmt.Sprintf("awscli-%s", node.Name)
 	v.Logger.Info("Creating a test pod on the hybrid node for pod identity add-on to access aws resources")
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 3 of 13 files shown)


test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +50,49 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: namespace,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "nginx",
+					Image: fmt.Sprintf("%s.dkr.ecr.%s.amazonaws.com/ecr-public/nginx/nginx:latest", constants.EcrAccounId, "us-west-2"),
```
g-gaston, line 71: can we use the same region as the cluster? the more cross region traffic we avoid the better

g-gaston: if not, can use public ecr?

test/e2e/addon/podidentityverify.go
```diff
@@ -3,26 +3,31 @@ package addon
...
 	podIdentityDaemonSet = "eks-pod-identity-agent-hybrid"
+	podIdentityToken     = "eks-pod-identity-token"
 )
 
 type VerifyPodIdentityAddon struct {
 	Cluster   string
 	K8S       *clientgo.Clientset
 	Logger    logr.Logger
 	EKSClient *eks.Client
+	NodeIP    string
```
g-gaston, line 30: would it make more sense to pass the node name instead?

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +50,49 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	// Ref: https://docs.aws.amazon.com/eks/latest/userguide/pod-id-how-it-works.html#pod-id-agent-pod
+	if pod, err := v.K8S.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{}); err != nil {
+		return err
+	} else {
+		for _, volumeMount := range pod.Spec.Containers[0].VolumeMounts {
+			if volumeMount.Name == podIdentityToken {
+				return nil
+			}
+		}
+	}
```
g-gaston, line 95: This is clever. However, this is kind of testing the implementation instead of testing the feature
Why don't we just try to make use of the aws creds that should be assigned to the pod? That would be closer to testing the feature e2e. Regardless of how the creds are mounted on the pod.

test/e2e/kubernetes/kubernetes.go
```diff
@@ -214,19 +214,26 @@ func CreateNginxPodInNode(ctx context.Context, k8s *kubernetes.Clientset, nodeNa
 		},
 	}
 
+	return CreateNginxPod(ctx, k8s, pod, logger)
+}
+
+func CreateNginxPod(ctx context.Context, k8s *kubernetes.Clientset, pod *corev1.Pod, logger logr.Logger) error {
```
g-gaston, line 220: Should this be named `CreatePod`? there is nothing particular about nginx in this method right?

test/e2e/addon/podidentityverify.go
jaxesn: Yeah i agree, i was kind of thinking of actually use the aws cli in a container and do like an s3 ls or something, wahtever permission we want to give it?

test/e2e/addon/podidentityverify.go
junpengdev: WaitForNode accepts nodeId as a parameter. I wanted to call `WatiForNode` to ensure node is ready.

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +71,170 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	bucketObjectKey := "test"
+	bucketObjectContent := "RANDOM-WORD"
+	if _, err := v.S3Client.PutObject(ctx, &s3.PutObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(bucketObjectKey),
+		Body:   strings.NewReader(bucketObjectContent),
+	}); err != nil {
+		return err
+	}
+	v.Logger.Info("Upload text file to S3 bucket", "key", bucketObjectKey, "content", bucketObjectContent)
```
g-gaston, line 89: Does this need to be done per test or can we do it only once as part of the setup?

test/e2e/cluster/cfn-templates/setup-cfn.yaml
```diff
@@ -362,6 +362,14 @@ Resources:
                 path: /root/download-private-key.sh
                 permissions: "0755"
 
+  PodIdentityS3Bucket:
+    Type: AWS::S3::Bucket
+    DeletionPolicy: Delete
+    Properties:
+      Tags:
+        - Key: !Ref TestClusterTagKey
+          Value: !Ref ClusterName
```
g-gaston, line 371: could we create the object in s3 from here as well so we don't have to handle it separetly? (`S3Object`)

example/hybrid-ira-cfn.yaml
```diff
@@ -48,7 +48,8 @@ Resources:
     Properties:
       RoleName: !Ref RoleName
       ManagedPolicyArns:
-        - 'arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly'
+        - arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly
+        - arn:aws:iam::aws:policy/AmazonEKSWorkerNodeMinimalPolicy
```
g-gaston, line 52: same here about example

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +71,170 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: namespace,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  podName,
+					Image: "public.ecr.aws/aws-cli/aws-cli",
```
jaxesn, line 108: if we are going to run this in the canary test, i think Id rather we pull from private ecr in the respective region. You can mirror this image just like we for nginx and the cni images.

junpengdev: Just curious why we want to pull it from private ecr? It is just performance concern?

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +71,170 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+					Command: []string{
+						"/bin/bash",
+						"-c",
+						"sleep 1h",
```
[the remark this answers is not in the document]
junpengdev: `sleep infinity` only works if it is implemented. It is not implemented in all linux distributions. I will change it to `sleep 86400` to sleep for a day.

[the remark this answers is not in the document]
jaxesn: well is it implemented in the nginx image we are using? if so, it doesnt matter that its not in other OSs

test/e2e/addon/podidentityverify.go
junpengdev: Moved this to cluster creation.

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +71,170 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	v.Logger.Info("Check output from exec pod with command", "output", stdout)
+
+	if stdout != bucketObjectContent {
+		return fmt.Errorf("failed to get object %s from S3 bucket %s", bucketObjectKey, bucket)
+	}
 
 	return nil
 }
+
+func getBucketNameFromPodIdentityAssociation(ctx context.Context, cluster string, eksClient *eks.Client, iamClient *iam.Client) (string, error) {
```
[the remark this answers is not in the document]
junpengdev: It's different than node ip or cluster name. For now pod identity S3 bucket is created per stack during cluster creation. The constraint for S3 bucket name is it has to be unique across all AWS accounts. I didn't see a straightforward way to pass the bucket name from cluster creation to test suite. I am open to any suggestions.

test/e2e/addon/podidentityaddon.go
```diff
@@ -59,3 +64,15 @@ func (p PodIdentityAddon) Create(ctx context.Context, logger logr.Logger, eksCli
 
 	return nil
 }
+
+func (p PodIdentityAddon) Upload(ctx context.Context, logger logr.Logger, client *s3.Client, bucket string) error {
```
g-gaston, line 68:
```suggestion
func (p PodIdentityAddon) UploadFileForVerification(ctx context.Context, logger logr.Logger, client *s3.Client, bucket string) error {
```

test/e2e/cluster/create.go
```diff
@@ -106,13 +108,19 @@ func (c *Create) Run(ctx context.Context, test TestResources) error {
...
 
-	podIdentityAddon := addon.NewPodIdentityAddon(hybridCluster.Name, podIdentityAddonName, stackOut.podIdentityRoleArn)
+	podIdentityAddon := addon.NewPodIdentityAddon(hybridCluster.Name, stackOut.podIdentity.roleArn)
 
 	err = podIdentityAddon.Create(ctx, c.logger, c.eks, k8sClient)
 	if err != nil {
 		return fmt.Errorf("creating add-on %s for EKS cluster: %w", podIdentityAddon.Name, err)
 	}
 
+	// upload test file to pod identity S3 bucket
```
g-gaston, line 118: yeah this reinforces my previous comment. instead of adding a comment to say what the method does, name the method after what it does

test/e2e/cluster/stack.go
```diff
@@ -281,6 +306,40 @@ func (s *stack) delete(ctx context.Context, clusterName string) error {
 	return nil
 }
 
+func emptyS3Bucket(ctx context.Context, client *s3.Client, bucket *string) error {
+	var err error
```
g-gaston, line 310:
```suggestion
```

test/e2e/cluster/stack.go
```diff
@@ -265,7 +275,22 @@ func waitForStackOperation(ctx context.Context, client *cloudformation.Client, s
 func (s *stack) delete(ctx context.Context, clusterName string) error {
 	stackName := stackName(clusterName)
 	s.logger.Info("Deleting E2E test cluster stack", "stackName", stackName)
-	_, err := s.cfn.DeleteStack(ctx, &cloudformation.DeleteStackInput{
+
+	var output *cloudformation.DescribeStackResourceOutput
```
g-gaston, line 279:
```suggestion
```
what's this for?

test/e2e/addon/podidentityverify.go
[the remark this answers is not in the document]
g-gaston: can you get the bucket based on the tag? the cluster name is unique and you are tagging the bucket with it. Descrive buckets using the tag as a filter? if we have more than one bucket where we are using the same tag, just add another that to this one like "pod identity: true" or whatever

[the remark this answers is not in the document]
jaxesn: We do something similiar for the jumpbox: https://github.com/aws/eks-hybrid/blob/main/test/e2e/peered/jumpbox.go

test/e2e/cluster/stack.go
```diff
@@ -281,6 +306,42 @@ func (s *stack) delete(ctx context.Context, clusterName string) error {
 	return nil
 }
 
+func emptyS3Bucket(ctx context.Context, client *s3.Client, bucket *string) error {
```
g-gaston, line 309: nit
why is this in the stack and not in `Delete.Run`?
it seems to match better the creation flow

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +69,95 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	for _, bucket := range listBucketsOutput.Buckets {
+		getBucketTaggingOutput, err := client.GetBucketTagging(ctx, &s3.GetBucketTaggingInput{
+			Bucket: bucket.Name,
+		})
+		if err != nil {
+			// Ignore if there is an error while retrieving tags
```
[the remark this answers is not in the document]
junpengdev: I am going to remove it. Previously it loops all s3 bucket but I only allow s3:GetBucketTagging on buckets with prefix `arn:aws:s3:::ekshybridci-arch-`. Now since I am using prefix `ekshybridci-arch-` to filter buckets in `ListBuckets`, this is not needed.

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +69,95 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+			}
+
+			if *tag.Key == "aws:cloudformation:logical-id" && *tag.Value == PodIdentityS3Bucket {
+				foundPodIdentityTag = true
+			}
+
+			if foundClusterTag && foundPodIdentityTag {
+				return *bucket.Name, nil
+			}
+		}
```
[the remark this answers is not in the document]
junpengdev: There should be only one match. If both tags found and matches, then it returns the bucket name.

test/e2e/cluster/stack.go
junpengdev: The reason I put it here is it interacts with CFN client on stack. Right now it gets the S3 bucket from stack name.

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +69,95 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	v.Logger.Info("Check output from exec pod with command", "output", stdout)
+
+	if stdout != bucketObjectContent {
+		return fmt.Errorf("failed to get object %s from S3 bucket %s", bucketObjectKey, bucket)
+	}
 
 	return nil
 }
+
+func getPodIdentityS3Bucket(ctx context.Context, cluster string, client *s3.Client) (string, error) {
```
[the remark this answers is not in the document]
junpengdev: moved to buildPeeredVPCTestForSuite

test/e2e/addon/podidentityverify.go
```diff
@@ -45,7 +69,95 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
...
+	listBucketsOutput, err := client.ListBuckets(ctx, &s3.ListBucketsInput{
+		Prefix: aws.String("ekshybridci-arch-"),
```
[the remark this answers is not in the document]
junpengdev: It is not well supported in cloud formation template. We can use something below to specify the prefix but then we don't have the guarantee that bucket name needs to be unique.

```
Properties: 
      Name: !Sub "${ClusterName}-PodIdentity-bucket"
```

test/e2e/addon/podidentityverify.go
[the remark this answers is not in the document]
g-gaston: what if one matches and the other doesn't?

test/e2e/addon/podidentityverify.go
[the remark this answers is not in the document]
g-gaston: What about using a configurable bucket name prefix? one that doesn't depend on the name of the stack

like
```yaml
BucketName: !Join 
  - '-'
  - - !Ref PodIdentityBucketPrefix
    - !Ref AWS::AccountId
    - !Ref AWS::Region
    - !Select 
      - 0
      - !Split 
        - '-'
        - !Select 
          - 2
          - !Split 
            - /
            - !Ref AWS::StackId
```

So we can pass `PodIdentityBucketPrefix` as a parameter to the stack. And reference that same constant from your code.

test/e2e/cluster/stack.go
g-gaston: we could use the same mechanism to get the bucket name we use in the test to get it
But sure, if this makes it easier, sounds good

test/e2e/suite/nodeadm_test.go
```diff
@@ -481,6 +484,14 @@ func buildPeeredVPCTestForSuite(ctx context.Context, suite *suiteConfiguration)
...
 	}
 
+	test.podIdentityS3Bucket, err = getPodIdentityS3Bucket(ctx, test.cluster.Name, test.s3Client)
+	if err != nil {
+		return nil, err
+	}
+
+	if test.podIdentityS3Bucket == "" {
+		return nil, fmt.Errorf("s3 bucket for pod identity not found")
+	}
```
g-gaston, line 494: shouldn't this be covered by checking the error in line 488?

test/e2e/suite/nodeadm_test.go
```diff
@@ -557,3 +573,37 @@ func newLoggerForTests() e2e.PausableLogger {
...
+			return "", err
+		}
+
+		var foundClusterTag, foundPodIdentityTag bool
+		for _, tag := range getBucketTaggingOutput.TagSet {
+			if *tag.Key == constants.TestClusterTagKey && *tag.Value == cluster {
+				foundClusterTag = true
+			}
+
+			if *tag.Key == "aws:cloudformation:logical-id" && *tag.Value == addon.PodIdentityS3Bucket {
```
g-gaston, line 599: can we add a tag ourselves instead of relying on the cfn tag?

test/e2e/addon/podidentityverify.go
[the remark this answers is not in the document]
junpengdev: Thanks for this suggestion! It worked.

test/e2e/suite/nodeadm_test.go
```diff
@@ -576,7 +574,7 @@ func newLoggerForTests() e2e.PausableLogger {
 
 func getPodIdentityS3Bucket(ctx context.Context, cluster string, client *s3v2.Client) (string, error) {
 	listBucketsOutput, err := client.ListBuckets(ctx, &s3v2.ListBucketsInput{
-		Prefix: aws.String("ekshybridci-arch-"),
+		Prefix: aws.String(strings.ToLower(addon.PodIdentityS3Bucket)),
```
g-gaston, line 577: let's avoid having `ToLower` in two places
if you need it in lower case, let's define it that way in the constant
if we need a second constant, that's ok

test/e2e/addon/podidentityverify.go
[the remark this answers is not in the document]
g-gaston: I don't get it. If only one bucket can have either tag, why do we need to check for both? if it's impossible for one bucket to have one tag but not the other, let's just only check one tag. Else, we need to reset the flags.

test/e2e/addon/podidentityverify.go
jaxesn: sorry, i though it had responded here. its mainly to be consistent with the other images we are pulling during these tests. We have the ecr mirrors setup in each region to isolate as much as we can between regions.

test/e2e/cluster/cfn-templates/setup-cfn.yaml
```diff
@@ -362,6 +366,31 @@ Resources:
...
                 permissions: "0755"
 
+  PodIdentityS3Bucket:
+    Type: AWS::S3::Bucket
+    DeletionPolicy: Delete
+    Properties:
+      BucketName:
+        Fn::Join:
+        - '-'
+        - - !Ref PodIdentityS3BucketPrefix
```
jaxesn, line 376: maybe a comment here, i cant easily follow what we are doign with the splits.

junpengdev: Yeah it's a comment before. It is suggested to use some prefix for pod identity S3 bucket.

test/e2e/addon/podidentityverify.go
```diff
@@ -3,26 +3,53 @@ package addon
...
 const (
 	getAddonTimeout      = 2 * time.Minute
-	podIdentityAgent     = "eks-pod-identity-agent"
 	podIdentityDaemonSet = "eks-pod-identity-agent-hybrid"
+	podIdentityToken     = "eks-pod-identity-token"
+	policyName           = "pod-identity-association-role-policy"
+	PodIdentityS3Bucket  = "PodIdentityS3Bucket"
 )
 
+var PodIdentityS3BucketPrefix = strings.ToLower(PodIdentityS3Bucket)
```
g-gaston, line 30: just define it as a constant pls, hardcode the value if you need t, it doesn't need to be derived

g-gaston (approved):

test/e2e/cluster/cfn-templates/setup-cfn.yaml
jaxesn: can we just add a comment in the cfn yaml here that just explains what this is doing?

test/e2e/addon/podidentityverify.go
junpengdev: The image doesn't exist in the private ECR yet. What process do we follow to make it available in the private registry across all regions?

g-gaston (approved):

PR #349, opened 2025-02-11
Pin containerd to version 1.7.*

achen1210:
*Issue #, if available:* 2889

*Description of changes:*

Pins containerd version to 1.7.* instead of having the package manager fetch the latest version available to match EKS upstream (https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2023/variables-default.json#L13)

*Testing (if applicable):*

Tested install, init, upgrade, uninstall flow on Ubuntu and RHEL with docker and distro (when appropriate) containerd sources.


jaxesn:
Did one of the OS's we support ship a 2.x package?

internal/packagemanager/packagemanager.go
```diff
@@ -36,6 +36,9 @@ const (
 	aptDockerRepoSourceFilePath = "/etc/apt/sources.list.d/docker.list"
 	yumDockerRepoSourceFilePath = "/etc/yum.repos.d/docker-ce.repo"
 
+	// pin containerd to same version as EKS
+	containerd_version = "1.7.*"
```
vignesh-goutham, line 40: we use camelCase as standard. So this should be `containerdVersion`

internal/packagemanager/packagemanager.go
```diff
@@ -203,13 +206,25 @@ func (pm *DistroPackageManager) updateDockerAptPackagesWithRetries(ctx context.C
...
+}
+
 // GetContainerd gets the Package
 // Satisfies the containerd source interface
 func (pm *DistroPackageManager) GetContainerd() artifact.Package {
 	packageName := containerdDistroPkgName
 	if pm.dockerRepo != "" {
 		packageName = containerdDockerPkgName
 	}
+	packageName = pm.appendContainerdVersion(packageName)
```
vignesh-goutham, line 227: Lets abstract this too and put it all under a function `getContainerdPackageNameWithVersion`

internal/packagemanager/packagemanager.go
achen1210: done

internal/packagemanager/packagemanager.go
achen1210: done

internal/packagemanager/packagemanager.go
```diff
@@ -36,6 +36,9 @@ const (
 	aptDockerRepoSourceFilePath = "/etc/apt/sources.list.d/docker.list"
 	yumDockerRepoSourceFilePath = "/etc/yum.repos.d/docker-ce.repo"
 
+	// pin containerd to same version as EKS
+	containerdVersion = "1.7.*"
```
g-gaston, line 40: curious why is this in the package manager package and not in the containerd package?

achen1210: The package manager is responsible for installing containerd so I put it here since the version will only matter during install; the containerd Source is abstracted away in the containerd package and package manager is responsible for implementing GetContainerd()

g-gaston: Doesn't that sound backwards to you? If we want to change the containerd version, the packagamanager package changes but the containerd one doesn't 

Another way to look at it: if you want to check what version of containerd we use and you know nothing about our codebase, where would you look first?

achen1210: Ok I see what you're saying, moved

jaxesn (approved):

internal/packagemanager/packagemanager.go
```diff
@@ -203,13 +203,29 @@ func (pm *DistroPackageManager) updateDockerAptPackagesWithRetries(ctx context.C
 	return cmd.Retry(ctx, pm.updateDockerAptPackagesCommand, 5*time.Second)
 }
 
-// GetContainerd gets the Package
-// Satisfies the containerd source interface
-func (pm *DistroPackageManager) GetContainerd() artifact.Package {
+func (pm *DistroPackageManager) appendContainerdVersion(target string) string {
+	switch pm.manager {
+	case yumPackageManager:
+		return fmt.Sprintf("%s-%s", target, containerd.ContainerdVersion)
```
g-gaston, line 209: why is this not an argument to appendContainerdVersion/GetContainerd?

achen1210: We can make this `func (pm *DistroPackageManager) appendPackageVersion(package string, version string) string {}` instead, and yeah we can add it to GetContainerd, I'll add it to one of my other changes

PR #350, opened 2025-02-11
Add node IP validation to nodeadm init WIP

achen1210:
*Issue #, if available:* 2892

*Description of changes:* 

A skippable error is added for the user if the node ip is not within the RemoteNetworkConfig CIDR blocks during nodeadm init, and the user is directed to set the node-ip in the kubelet flags section of the config file. An out of range node IP will prevent some features from working such as auto cert renewal.

*Testing (if applicable):*

Tested on Ubuntu.


internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,238 @@
...
+		var addrs []net.IP
+		var nodeName string
+		// If using SSM, the node name will be set at initialization to the SSM instance ID,
+		// so it won't resolve to anything via DNS, hence we're only checking in the case of IAM-RA
+		nodeName, err = getNodeName(IAMNodeName, kubeletArgs)
+		if err != nil {
+			return nil, err
+		}
+
+		addrs, _ = net.LookupIP(nodeName)
```
jaxesn, line 217: what happens here if the nodename is empty and/or does not resolve to an ip?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,238 @@
...
+	debugCidrs := getClusterCIDRs(hnp.remoteNetworkConfig)
+	fmt.Printf("DEBUG: cidrs:  %v\n", debugCidrs)
+
+	if validIP, err := isIPInClusterNetworks(ipAddr, hnp.remoteNetworkConfig); err != nil {
+		return err
+	} else if !validIP {
+		cidrs := getClusterCIDRs(hnp.remoteNetworkConfig)
+
+		return fmt.Errorf(
+			"node IP %s is not in any of the remote network CIDR blocks: %s; "+
```
jaxesn, line 60: we should discuss the exact message here with the team. i like pointing the user to use node-ip, but want to make sure we are clear as possible.

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,238 @@
...
+				"or use --skip ip-validation",
+			ipAddr, cidrs,
+		)
+	}
+	return nil
+}
+
+func getClusterCIDRs(remoteNetworkConfig *eks.RemoteNetworkConfig) []string {
+	var cidrs []string
+	for _, network := range remoteNetworkConfig.RemoteNodeNetworks {
```
jaxesn, line 72: what happens if we do have perms to get the remotenodenetworks? do we just skip this new check altogether?

internal/node/hybrid/ip_validator.go
achen1210: This part is going to be re-done - the original idea is that if the nodename is empty there will be an error since even if there is no IAMNodeName and no hostname-override set by kubelet flags, the nodeutil.GetHostname(hostnameOverride) will get a non-empty hostname (defaulting to kernel reported hostname if necessary) or throw an error if it's unable to - see https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-helpers/node/util/hostname.go#L28. This is also how kubelet will get the determine the nodeName (https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L1228) - we do not have a cloud provider for hybrid (we set "cloud-provider" flag to "") so the kubelet will default to the hostname coming out from this function as the nodeName.

However,  for hybrid nodes we always set hostname-override to cfg.Status.Hybrid.NodeName, which is either the SSM instance ID or the IAM-RA NodeName specified in .spec.hybrid.iamRolesAnywhere.nodeName. Although the user could overwrite this using kubelet flags in the nodeConfig, in practice the user cannot set a different hostname since the node will be unable to join the cluster as the name would not match the name in the credential provider. For SSM, as the instance ID is generated during registration, we can assume the DNS will not resolve to anything; however for IAM-RA the user may have a DNS entry for their node name in their network. Hence for this portion we only need to check for the .spec.hybrid.iamRolesAnywhere.nodeName in case they're using IAM-RA, and we don't need to use the nodeutil.GetHostname() function or checking the kubelet flags.

If the DNS does not resolve to an IP kubelet uses the ResolveBindAddress function to get an IP from the available default routes.

internal/node/hybrid/ip_validator.go
achen1210: During enrich config prior to this validation we check that RemoteNetworkConfig was read during DescribeCluster and isn't nil (https://github.com/aws/eks-hybrid/blob/main/internal/node/hybrid/configenricher.go#L61), if the RemoteNetworkConfig is able to be read then I think it always gets the whole thing including the remote node networks. But we can add a check and error out in case, and the user can use --skip if necessary

jaxesn: yeah i wanted to make sure we match whatever we do today when checking if there are remotenodeconfig for this new functionality as well. so if the user doesnt give us permissions, we just skip over it (assuming thats what we do today)

achen1210: yeah sounds good right now we require users to have remoteNetworkConfig enabled on their EKS cluster to use hybrid nodes so by this point we can assume the remoteNetworkConfig exists - i'll double check on some of the edge cases though and see if it can be missing remote node networks etc.

PR #351, opened 2025-02-13
adds gpg verification to ssm-setup-cli download

jaxesn:
*Issue #, if available:*

*Description of changes:*

When downloading the ssm setup cli, this adds gpg verification based on the SSM team's public key and signature file that is released along with the artifacts.  

The ssm agent uses these [signature](https://github.com/aws/amazon-ssm-agent/blob/mainline/agent/setupcli/managers/verificationmanagers/linuxmanager.go#L47) files when install/updating itself.  The implementation there exec's out to gpg on the host to do the verification, basically following [these](https://docs.aws.amazon.com/systems-manager/latest/userguide/verify-agent-signature.html) steps from their docs.  This could be an alternative approach we could follow instead of what I have done here.

This approach uses gopenpgp from [ProtonMail](https://github.com/ProtonMail/gopenpgp). The go stdlib includes [opengpg](https://pkg.go.dev/golang.org/x/crypto/openpgp) as part of the x/crypto package, however it has been deprecation due to lack of ownership.  ProtonMail has forked this lib and built goopenpgp around it.  This feels cleaner to me than exec-ing out to gpg, but open to thoughts.

*Testing (if applicable):*

Verified the integration tests execute this change. Manually "broke" the signature to confirm it will fail validation.  Happy to add some unit tests here, but the relevant code does not have any yet, so I could use some assistance getting started.


internal/ssm/install.go
```diff
@@ -19,13 +21,46 @@ import (
 )
 
 const (
-	installerPath = "/opt/ssm/ssm-setup-cli"
-	configRoot    = "/etc/amazon"
+	installerPath   = "/opt/ssm/ssm-setup-cli"
```
jaxesn, line 24: this key is in the progress of being rotated, it expires next week. Ill wait to merge this PR until we get a new release of ssm using the new key for the signature.

jaxesn: I was also thinking it may make sense to add this to our nodeadm manifest instead of built into the code. Open to thoughts here?

g-gaston: i vote to have it here for now
i like the idea of having it in the manifest but, that doesn't really work well (from a security pov) until we sign the manifest itself.

internal/ssm/install.go
```diff
@@ -34,21 +80,79 @@ type PkgSource interface {
...
 	}
+	defer signature.Close()
 
-	if err = runInstallWithRetries(ctx); err != nil {
-		return errors.Wrapf(err, "failed to install ssm agent")
+	var installerBuffer bytes.Buffer
+	installerTee := io.TeeReader(installer, &installerBuffer)
+
+	if err := validateSetupSignature(installerTee, signature); err != nil {
+		return err
```
g-gaston, line 123:
```suggestion
		return fmt.Errorf("validating ssm-setup-cli signature: %w, err)
```

i think here the context is super important, if not, we are just gonna get the error from the pgp library without telling us what's the signature being valdating

internal/ssm/source.go
```diff
@@ -45,6 +45,19 @@ func (s ssmInstallerSource) GetSSMInstaller(ctx context.Context) (io.ReadCloser,
...
 }
 
+func (s ssmInstallerSource) GetSSMInstallerSignature(ctx context.Context) (io.ReadCloser, error) {
+	endpoint, err := buildSSMURL(s.region)
+	if err != nil {
+		return nil, err
+	}
+	obj, err := util.GetHttpFileReader(ctx, endpoint+".sig")
+	if err != nil {
+		obj.Close()
```
g-gaston, line 55: does the obj need to be closed when there is an error? maybe we need to check the implementation, but generally when a "constructor" builds something that needs to be closed, that thing should only be closed when the constructor succeeds. Else, you might be. it might blow up since you could be getting a nil object

internal/ssm/source.go
```diff
@@ -45,6 +45,19 @@ func (s ssmInstallerSource) GetSSMInstaller(ctx context.Context) (io.ReadCloser,
 	return obj, nil
 }
 
+func (s ssmInstallerSource) GetSSMInstallerSignature(ctx context.Context) (io.ReadCloser, error) {
```
g-gaston, line 48: how difficult would it be to test this method?

internal/ssm/install.go
jaxesn: yeah we can add it the manifest once we get signing in.

internal/ssm/install_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
+			}
+
+			assert.NoError(t, err)
+
+			installedData, err := os.ReadFile(ssm.InstallerPath)
+			assert.NoError(t, err)
```
g-gaston, line 126: could we use gomega? nothing against testify but it seems like gomega is more used in this repo?

internal/ssm/install_test.go
```diff
@@ -0,0 +1,132 @@
...
+			ssm.InstallerPath = filepath.Join(tmpDir, "ssm-setup-cli")
+
+			server := test.NewHTTPServer(t, func(w http.ResponseWriter, r *http.Request) {
+				if tt.statusCode != http.StatusOK {
+					w.WriteHeader(tt.statusCode)
+					return
+				}
+
+				if r.URL.Path == "/latest/linux_amd64/ssm-setup-cli" {
+					w.Write(tt.installerData)
```
g-gaston, line 100: check for error from write and write an error code if so?

internal/ssm/install_test.go
```diff
@@ -0,0 +1,132 @@
...
+	signature, err := signer.Sign(data, crypto.Bytes)
+	assert.NoError(t, err)
+	return signature
+}
+
+func TestInstall(t *testing.T) {
+	publicKey, privateKey := generateKeyPair(t)
+	_, wrongPrivateKey := generateKeyPair(t)
+
+	installerData := []byte("#!/bin/bash\necho 'test installer'")
```
g-gaston, line 45: do we need this? it's ok, but i don't love making the tests depend on having bash available

I think an empty file should work?

internal/ssm/install_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	tests := []struct {
+		name          string
+		installerData []byte
+		signature     []byte
+		statusCode    int
+		wantErr       bool
```
g-gaston, line 52: a more robust way to do this is to have here the error message as a string (or part of the message)

That way not only you test that you get an error when you should, you check that you get the error that you expect getting

internal/ssm/install.go
```diff
@@ -18,37 +21,99 @@ import (
 	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
-const (
-	installerPath = "/opt/ssm/ssm-setup-cli"
+// TODO: a better way to override this
+var (
+	InstallerPath = "/opt/ssm/ssm-setup-cli"
```
g-gaston, line 26: i think there are two ways do avoid this
* make the path configurable in the function we call: either adding some kind of options to the install method (variadic clousures?) or making this function a method and configuring the path as a member of the receiver 
* Create a new method `InstallTo` that takes a path. From the tests we would test this method, and we would change `Install` to call `Install(..., installerPath)`. This doesn't test "all the code" but hides details from the caller if it doesn't need to configure the path.

I think i lean towards the latter but your call

internal/ssm/install.go
```diff
@@ -18,37 +21,99 @@ import (
...
+
+	verifyDataReader, err := verifier.VerifyingReader(installer, signature, crypto.Bytes)
+	if err != nil {
+		return err
+	}
+	verifyResult, err := verifyDataReader.ReadAllAndVerifySignature()
+	if err != nil {
+		return err
+	}
+	if sigErr := verifyResult.SignatureError(); sigErr != nil {
```
g-gaston, line 113: nit, do we need a special name for the err?
```suggestion
	if err := verifyResult.SignatureError(); err != nil {
```

internal/ssm/source.go
```diff
@@ -17,42 +17,128 @@ import (
...
+VplOoNgGnN19i3nuMcPf2npFICJv7kTIyn5Fh7pjWDCahl3U/PwoLjrrlEzpyStU
+oXSZrK3kiAADEdSODXJl8KYU0Pb27JbRr1ZbWnxb+O39TOhtssstulkR0v+IDGDQ
+rQE1b12sKgcNFSzInzWrNGu4S06WN8DYzlrTZ9aSHj+37ZqpXAevi8WOFXKPV3PA
+E6+O8RI2451Dcg==
+=aDkv
+-----END PGP PUBLIC KEY BLOCK-----`
+
+type SSMInstallerOption func(*ssmInstallerSource)
+
+// WithURLBuilder allows overriding the SSM URL building logic
```
g-gaston, line 64:
```suggestion
// WithURLBuilder allows overriding the SSM installer download URL.
```

internal/ssm/install.go
```diff
@@ -18,37 +21,104 @@ import (
 	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
-const (
-	installerPath = "/opt/ssm/ssm-setup-cli"
-	configRoot    = "/etc/amazon"
+// TODO: a better way to override this
+var (
```
g-gaston, line 25: i think this can go back to a const?
```suggestion
const (
```

internal/ssm/install.go
```diff
@@ -18,37 +21,104 @@ import (
...
+	PublicKey() string
 }
 
 // PkgSource serves and defines the package for target platform
 type PkgSource interface {
 	GetSSMPackage() artifact.Package
 }
 
-func Install(ctx context.Context, tracker *tracker.Tracker, source Source) error {
+func InstallTo(ctx context.Context, tracker *tracker.Tracker, source Source, logger *zap.Logger, installerPath string) error {
```
g-gaston, line 42: we should prob think to group the last 4 in a struct, 5 arguments is starting to be too high

g-gaston (approved):

g-gaston (approved):

PR #358, opened 2025-02-19
Node IP Validation

achen1210:
*Issue #, if available:* 2892

*Description of changes:* 

A skippable error is added for the user if the node ip is not within the RemoteNetworkConfig CIDR blocks during nodeadm init, and the user is directed to set the node-ip in the kubelet flags section of the config file. An out of range node IP will prevent some features from working such as auto cert renewal.

*Testing (if applicable):*

Tested on Ubuntu and with e2e tests.


internal/node/hybrid/hybrid.go
```diff
@@ -50,6 +61,36 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
...
+func (hnp *HybridNodeProvider) ValidateNodeIP(ctx context.Context) error {
+	// For hybrid nodes, we don't set the --node-ip flag anywhere else,
+	// so we can directly check if user has specified it in the config file
+	kubeletArgs := hnp.nodeConfig.Spec.Kubelet.Flags
+	var IAMNodeName string
+	if hnp.nodeConfig.IsIAMRolesAnywhere() {
+		IAMNodeName = hnp.nodeConfig.Spec.Hybrid.IAMRolesAnywhere.NodeName
+	}
+
+	// If remoteNetworkConfig is nil we didn't call ReadCluster before so we do that now
```
g-gaston, line 73: i don't think this is goodWe should not be calling this API unless the user doesn't provide all the necessary values in the cluster spec. That's the expected behavior. If not, there is no way to avoid calls to the eks API, which some users might want to avoid.

internal/node/hybrid/hybrid.go
```diff
@@ -50,6 +61,36 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
...
 }
 
+func (hnp *HybridNodeProvider) ValidateNodeIP(ctx context.Context) error {
+	// For hybrid nodes, we don't set the --node-ip flag anywhere else,
+	// so we can directly check if user has specified it in the config file
+	kubeletArgs := hnp.nodeConfig.Spec.Kubelet.Flags
+	var IAMNodeName string
+	if hnp.nodeConfig.IsIAMRolesAnywhere() {
+		IAMNodeName = hnp.nodeConfig.Spec.Hybrid.IAMRolesAnywhere.NodeName
+	}
```
g-gaston, line 71: why can't we rely on the node name set in the status?

internal/node/hybrid/hybrid.go
```diff
@@ -50,6 +61,36 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
 	return hnp.logger
 }
 
+func (hnp *HybridNodeProvider) ValidateNodeIP(ctx context.Context) error {
+	// For hybrid nodes, we don't set the --node-ip flag anywhere else,
+	// so we can directly check if user has specified it in the config file
+	kubeletArgs := hnp.nodeConfig.Spec.Kubelet.Flags
+	var IAMNodeName string
```
g-gaston, line 68:
```suggestion
	var iamNodeName string
```

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return fmt.Errorf("node IP: %q not found in the host's network interfaces", nodeIP.String())
+}
+
+// getNodeIP determines the node's IP address based on kubelet configuration and system information.
+func getNodeIP(kubeletArgs []string, IAMNodeName string) (net.IP, error) {
+	// Follows algorithm used by kubelet to assign nodeIP
+	// Implementation adapted for hybrid nodes
+	// 1) Use nodeIP if set (and not "0.0.0.0"/"::")
+	// 2) If the user has specified an IP to HostnameOverride, use it
```
g-gaston, line 173: do we even allow this? i don't think this will work right? wouldn't that ip become the node name?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		if ip != nil && ip.Equal(nodeIP) {
+			return nil
+		}
+	}
+	return fmt.Errorf("node IP: %q not found in the host's network interfaces", nodeIP.String())
+}
+
+// getNodeIP determines the node's IP address based on kubelet configuration and system information.
+func getNodeIP(kubeletArgs []string, IAMNodeName string) (net.IP, error) {
```
g-gaston, line 169: is it possible to set any of this configuration through the kubelet config file? bc i believe we do allow users to provide a patch for this file that we merge with our own config

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,226 @@
...
+	// Follows algorithm used by kubelet to assign nodeIP
+	// Implementation adapted for hybrid nodes
+	// 1) Use nodeIP if set (and not "0.0.0.0"/"::")
+	// 2) If the user has specified an IP to HostnameOverride, use it
+	// 3) Lookup the IP from node name by DNS
```
g-gaston, line 174: do we need this?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,226 @@
...
+	var ipAddr net.IP
+
+	nodeIPSpecified := nodeIP != nil && nodeIP.To4() != nil && !nodeIP.IsUnspecified()
+
+	if nodeIPSpecified {
+		ipAddr = nodeIP
+	} else if parsedAddr := k8snet.ParseIPSloppy(hostname); parsedAddr != nil {
+		if parsedAddr.To4() == nil {
+			return nil, fmt.Errorf("hostname address %s is not IPv4", parsedAddr)
+		}
```
g-gaston, line 196: what's the implication of this? not against it but just trying to understand the details
How does this behave if the node has both ipv4 and ipv6 ips?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,226 @@
...
+
+		if ipAddr == nil {
+			// current standard function for resolving bind address: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L768
+			ipAddr, err = apimachinerynet.ResolveBindAddress(nodeIP)
+		}
+
+		if ipAddr == nil {
+			// We tried everything we could, but the IP address wasn't fetchable; error out
+			return nil, fmt.Errorf("couldn't get ip address of node: %w", err)
```
g-gaston, line 220: i wonder if we should treat this a best effort and not error out
it's the users error that we can't determine the IP to validate. Should that be an error that they need to handle?

internal/node/hybrid/ip_validator_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/aws/eks-hybrid/internal/aws/eks"
+	"github.com/aws/eks-hybrid/internal/node/hybrid"
+)
+
+func TestHybridNodeProvider_ValidateNodeIP(t *testing.T) {
+	tests := []struct {
+		name                string
+		nodeConfig          *api.NodeConfig
+		remoteNetworkConfig *eks.RemoteNetworkConfig
+		wantErr             bool
```
g-gaston, line 22: i would suggest making this a string so you can actually expect the error message (or partial error message)

internal/nodeprovider/interface.go
```diff
@@ -37,6 +37,9 @@ type NodeProvider interface {
 	// Logger defines the logger for the node provider
 	Logger() *zap.Logger
 
+	// ValidateNodeIP validates the node ip is within the allowed range
```
g-gaston, line 40: nit this doesn't make sense in the context of a provider interface. the concept of allowed range only applies to hybrid nodes, not ec2

what do you think about just `Validate(ctx)`
there is not reason why need to expose in this interface what we validate
And this gives place to any other future validation we might want to make

internal/system/swap.go
```diff
@@ -134,7 +134,7 @@ type swap struct {
 	swapType string
 }
 
-// Mount represets the filesystem info
+// Mount represents the filesystem info
```
g-gaston, line 137:
```suggestion
// mount represents the filesystem info
```

internal/node/hybrid/configenricher.go
```diff
@@ -35,21 +35,26 @@ func (p *HybridNodeProvider) Enrich(ctx context.Context) error {
...
 	return cluster.Cluster, nil
 }
 
-func ensureClusterDetails(ctx context.Context, awsConfig aws.Config, nodeConfig *api.NodeConfig) error {
-	cluster, err := readCluster(ctx, awsConfig, nodeConfig)
+func needsClusterDetails(nodeConfig *api.NodeConfig) bool {
+	return nodeConfig.Spec.Cluster.APIServerEndpoint == "" || nodeConfig.Spec.Cluster.CertificateAuthority == nil || nodeConfig.Spec.Cluster.CIDR == ""
+}
+
+func ensureClusterDetails(ctx context.Context, p *HybridNodeProvider) error {
```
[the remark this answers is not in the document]
achen1210: it was to save the remotenetworkconfig into hybridnodeprovider if we've already called describecluster so we don't need to call it again, if we think an extra call is not a problem or we see a better way this can be removed/changed

internal/node/hybrid/configenricher.go
```diff
@@ -35,21 +35,26 @@ func (p *HybridNodeProvider) Enrich(ctx context.Context) error {
...
+// ReadCluster calls eks.DescribeCluster and returns the cluster after ensuring it has a RemoteNetworkConfig
+func ReadCluster(ctx context.Context, awsConfig aws.Config, nodeConfig *api.NodeConfig) (*eks.Cluster, error) {
 	cluster, err := eks.DescribeCluster(ctx, eks.NewClient(awsConfig), nodeConfig.Spec.Cluster.Name)
 	if err != nil {
 		return nil, err
 	}
 
+	if cluster.Cluster.RemoteNetworkConfig == nil {
+		return nil, errors.New("eks cluster does not have remoteNetworkConfig enabled, which is required for Hybrid Nodes")
+	}
```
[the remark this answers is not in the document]
achen1210: yeah this can be moved to a post readcluster validation function/step

internal/node/hybrid/hybrid.go
achen1210: yeah this can be replaced with hnp.nodeConfig.Status.Hybrid.NodeName

internal/node/hybrid/ip_validator.go
achen1210: i'm only aware of https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-nodeadm.html#_optional_kubelet_configuration which is where this function will get kubelet configuration details from

internal/node/hybrid/ip_validator.go
achen1210: yeah technically this is not needed because that ip would become the node name which wouldn't work for ssm and would have to match the iam-ra node name which we're already checking for earlier. this was to match the kubelet algorithm exactly to avoid any mistakes but we can take it out for hybrid nodes

internal/node/hybrid/ip_validator.go
achen1210: I think it's possible for them to have a dns entry set for their node for other purposes (only in the case of iam-ra, since for ssm node name is set to instance id) and a misconfigured ip in the dns server could cause issues when using it as a hybrid node

internal/node/hybrid/ip_validator.go
achen1210: this is for the hostname which we can remove, the idea was to prevent setting a ipv6 node name and then having the node ip be ipv6 and cause some unknown error down the line. this is going to be an error either way for the user since kubelet would use the ipv6 address as the node ip and then they would have issues since hybrid nodes wouldn't support it.

internal/node/hybrid/ip_validator.go
achen1210: this is from kubelet's code so it would have same error when initializing, maybe there's some case where they want to run init but not configure or run kubelet yet so we could consider exiting gracefully from the validation. It also would be really unlikely this ever happens though I think, there would have to be no network interfaces available and so there's something really wrong with their node

internal/nodeprovider/interface.go
achen1210: yeah this sounds good

internal/node/hybrid/hybrid.go
achen1210: changed: we skip validation with a warning message if describecluster was not yet called, and error if describecluster was called and validation fails

internal/node/hybrid/hybrid.go
```diff
@@ -50,7 +63,59 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
...
 }
 
+func (hnp *HybridNodeProvider) Validate(ctx context.Context, skipPhases []string) error {
+	if !slices.Contains(skipPhases, ipValidation) {
+		if hnp.cluster == nil {
+			hnp.Logger().Info("Warning: EKS Cluster details not retrieved - IP validation skipped")
+		} else {
+			hnp.logger.Info("Validating Node IP...")
+
+			// Only check flags set by user since hybrid nodes do not set --node-ip flag
```
jaxesn, line 73: what does this mean?

internal/node/hybrid/hybrid.go
```diff
@@ -50,7 +63,59 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
...
+			var iamNodeName string
+			if hnp.nodeConfig.IsIAMRolesAnywhere() {
+				iamNodeName = hnp.nodeConfig.Status.Hybrid.NodeName
+			}
+			nodeIp, err := getNodeIP(kubeletArgs, iamNodeName)
+			if err != nil {
+				return err
+			}
+
+			cluster, err := hnp.Cluster(ctx)
```
jaxesn, line 85: since we are checking for nil above, will we ever call readCluster?

test/integration/cases/hybrid-node-system-aspects/run.sh
```diff
@@ -22,7 +22,7 @@ systemctl start firewalld
 # allow cilium vxlan
 firewall-cmd --permanent --add-port=4789/udp
 
-nodeadm init --skip run,install-validation --config-source file://config.yaml
+nodeadm init --skip run,install-validation,ip-validation --config-source file://config.yaml
```
jaxesn, line 25: why do we have to skip these in the integration tests?

internal/node/hybrid/ip_validator.go
g-gaston: yeah but you are only getting the flags, right? is it possible to configure the same thing through `spec.kubelet.config`?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,221 @@
...
+func getNodeIP(kubeletArgs []string, iamNodeName string) (net.IP, error) {
+	// Follows algorithm used by kubelet to assign nodeIP
+	// Implementation adapted for hybrid nodes
+	// 1) Use nodeIP if set (and not "0.0.0.0"/"::")
+	// 2) If the user has specified an IP to HostnameOverride, use it (not allowed for hybrid nodes)
+	// 3) Lookup the IP from node name by DNS
+	// 4) Try to get the IP from the network interface used as default gateway
+	// Source: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/nodestatus/setters.go#L206
+
+	if err := ensureNoHostnameOverride(kubeletArgs); err != nil {
```
g-gaston, line 163: this is a good idea but i think it's in the wrong place
this method is supposed to give you the ip that kubelet will select for the node
but here we are running a validation specific for eks-hybrid

If so, this probably belong at the Validate or or even ValidateConfig level

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,221 @@
...
+		for _, cidr := range network.CIDRs {
+			if cidr != nil {
+				cidrs = append(cidrs, *cidr)
+			}
+		}
+	}
+	return cidrs
+}
+
+func extractFlagValue(kubeletArgs []string, flag string) (string, error) {
```
g-gaston, line 58: this method never returns an error, can we simplify?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,221 @@
...
+
+	if nodeIPSpecified {
+		ipAddr = nodeIP
+	} else {
+		// If using SSM, the node name will be set at initialization to the SSM instance ID,
+		// so it won't resolve to anything via DNS, hence we're only checking in the case of IAM-RA
+		if iamNodeName != "" {
+			var addrs []net.IP
+
+			addrs, _ = net.LookupIP(iamNodeName)
```
g-gaston, line 184: any reason to ignore the error?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,221 @@
...
+
+	nodeIPSpecified := nodeIP != nil && nodeIP.To4() != nil && !nodeIP.IsUnspecified()
+
+	if nodeIPSpecified {
+		ipAddr = nodeIP
+	} else {
+		// If using SSM, the node name will be set at initialization to the SSM instance ID,
+		// so it won't resolve to anything via DNS, hence we're only checking in the case of IAM-RA
+		if iamNodeName != "" {
+			var addrs []net.IP
```
g-gaston, line 182: any reason to define the variable here instead of using `:=`?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		if ip != nil && ip.Equal(nodeIP) {
+			return nil
+		}
+	}
+	return fmt.Errorf("node IP: %q not found in the host's network interfaces", nodeIP.String())
+}
+
+// getNodeIP determines the node's IP address based on kubelet configuration and system information.
+func getNodeIP(kubeletArgs []string, iamNodeName string) (net.IP, error) {
```
g-gaston, line 154: do we need to call this iamNodeName? to this method, it doesn't matter what time of node name
Only if the node name is know or not, right?

internal/node/hybrid/hybrid.go
achen1210: we can check the nodeconfig directly for the flags instead of using the flags in the kubelet struct (https://github.com/aws/eks-hybrid/blob/main/internal/kubelet/daemon.go#L23) which nodeadm adds some things to during init (i.e. setting hostname-override to ssm instance id). ultimately the kubelet cli command will have flags from this map + the flags from the user, but we want to do the validation before the kubelet daemon is configured and this map is made, which is ok since nodeadm doesn't add an node-ip flag on its own. will try to make comment more clear

internal/node/hybrid/hybrid.go
```diff
@@ -50,7 +63,59 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
...
+	return nil
+}
+
 func (hnp *HybridNodeProvider) Cleanup() error {
 	hnp.daemonManager.Close()
 	return nil
 }
+
+// Cluster retrieves the eks.Cluster object or makes a DescribeCluster call to the EKS API and caches the result if not already present
+func (p *HybridNodeProvider) Cluster(ctx context.Context) (*eks.Cluster, error) {
```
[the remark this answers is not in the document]
achen1210: should be hnp, will make the change. we called it p in configenricher and I didn't notice the inconsistency, will change it there too

internal/node/hybrid/hybrid.go
achen1210: no it'll only grab the cached cluster from the hnp struct. we can change this to hnp.cluster if it's more clear

internal/node/hybrid/ip_validator.go
achen1210: yeah this can be just nodeName

internal/node/hybrid/ip_validator.go
achen1210: no, it's a relic from the kubelet code but can clean it up

internal/node/hybrid/ip_validator.go
achen1210: this part is from the kubelet code for the way it sets the ip, and we just want to replicate it

test/integration/cases/hybrid-node-system-aspects/run.sh
achen1210: we currently can't call describecluster to get the remotenetworkconfig in the integration tests, when it's available in moto there'll be an integration tests that tests the ip-validation feature or we can add the mocked describecluster call

internal/node/hybrid/ip_validator_test.go
```diff
@@ -0,0 +1,161 @@
...
+			g := NewWithT(t)
+
+			hnp, err := hybrid.NewHybridNodeProvider(
+				tt.nodeConfig,
+				zap.NewNop(),
+				hybrid.WithCluster(tt.cluster),
+			)
+			g.Expect(err).To(Succeed())
+
+			err = hnp.Validate(context.Background(), []string{})
```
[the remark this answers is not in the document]
achen1210: will add

cmd/nodeadm/init/init.go
```diff
@@ -38,7 +38,7 @@ func NewInitCommand() cli.Command {
 	init.cmd = flaggy.NewSubcommand("init")
 	init.cmd.String(&init.configSource, "c", "config-source", "Source of node configuration. The format is a URI with supported schemes: [file, imds].")
 	init.cmd.StringSlice(&init.daemons, "d", "daemon", "Specify one or more of `containerd` and `kubelet`. This is intended for testing and should not be used in a production environment.")
-	init.cmd.StringSlice(&init.skipPhases, "s", "skip", "Phases of the bootstrap to skip. Allowed values: [install-validation, cni-validation, preprocess, config, run].")
+	init.cmd.StringSlice(&init.skipPhases, "s", "skip", "Phases of the bootstrap to skip. Allowed values: [install-validation, cni-validation, ip-validation, preprocess, config, run].")
```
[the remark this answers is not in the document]
achen1210: yeah this is more specific

internal/flows/init.go
```diff
@@ -37,6 +37,10 @@ func (i *Initer) Run(ctx context.Context) error {
 		return err
 	}
 
+	if err := i.NodeProvider.Validate(ctx, i.SkipPhases); err != nil {
```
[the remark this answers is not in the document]
achen1210: which constructor are you referring to for this? NodeProvider is an interface and it's already in initer; however user may want to skip some validations but not others so the validate function needs to have access to the skipphases

internal/node/hybrid/hybrid.go
```diff
@@ -50,7 +63,59 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
 	return hnp.logger
 }
 
+func (hnp *HybridNodeProvider) Validate(ctx context.Context, skipPhases []string) error {
+	if !slices.Contains(skipPhases, ipValidation) {
```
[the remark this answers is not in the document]
achen1210: I don't think I want to return early here since if there are other validations they would also be skipped. which branches are you referring to?

internal/node/hybrid/hybrid.go
```diff
@@ -50,7 +63,59 @@ func (hnp *HybridNodeProvider) Logger() *zap.Logger {
...
+		if hnp.cluster == nil {
+			hnp.Logger().Info("Warning: EKS Cluster details not retrieved - IP validation skipped")
```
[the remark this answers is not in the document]
achen1210: yeah this makes sense

internal/flows/init.go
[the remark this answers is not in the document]
g-gaston: the constructor that builds the node provider

internal/node/hybrid/ip_validator.go
g-gaston: What does this mean, that we ignore the error because the kubelet ignores the error?

If so, then i think we want to understand why the kuebelet ignores it, not just blindly copy it

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,221 @@
...
+				if err = validateNodeIP(addr); addr.To4() != nil && err == nil {
+					ipAddr = addr
+					break
+				}
+			}
+		}
+
+		if ipAddr == nil {
+			// current standard function for resolving bind address: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L768
+			ipAddr, err = apimachinerynet.ResolveBindAddress(nodeIP)
```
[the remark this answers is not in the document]
g-gaston: Pls don't just tell me this is what the kubelet does
I understand we are basing our logic on theirs, but this code is in your PR, you are writing it, so you should have a reason to write in this way

On the "if its an error then ipAddr will be nil", then just check the error instead of nil address, there is a reason the method returns an error as well. if you want to check both because you think you get nil ip and no error, that's ok. But the error gives more information than a nil address.

internal/node/hybrid/hybrid.go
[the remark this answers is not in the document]
g-gaston: what other validations? are there any?
I don't think we should write this method to accommodate for code that doesn't exist yet. It's easy to change this code later if add more validations.

But if you think we should, then let's do it properly. If there were more than one validation, then it would probably be too much logic for one method, so you would move the validations to multiple unexported methods and handle the orchestration from here.

If so, then you can write code like:

```go
if !slices.Contains(skipPhases, ipValidation) {
    validateNodeIP(nodeConfig, whatever)
}

if !slices.Contains(skipPhases, otherValidationName) {
   validateThisOtherThing()
}
```

Which is easier to read that the multiple nesting branching you have right now.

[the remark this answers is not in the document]
achen1210: yeah makes sense I think this is already implemented in the changes and you were looking at an old version

internal/node/hybrid/ip_validator.go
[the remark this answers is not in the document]
g-gaston: can we just do `if err != nil {` instead of `if ipAddr == nil {`?
and if you want to check both: `if err != nil || ipAddr == nil {`
I don't think the code needs to be identical to the kubelet's code

internal/flows/init.go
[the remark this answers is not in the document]
achen1210: i see, we can add this to the hybridnodeprovider and ec2nodeprovider structs, but it feels a little out of place - is there something wrong with the current approach?

internal/node/hybrid/ip_validator.go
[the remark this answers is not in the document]
achen1210: yeah thats fine since it looks like in resolvebindaddress() it will never return a valid ip in any of the cases where there's an error

however I think it's written this way in kubelet intentionally since it should attempt to use any ip it can get, even if theres an error during the retrieval process, and the real error is only if it can't get an ip. maybe it's possible that a change to resolvebindaddress or something in the future would cause a situation where kubelet would've been able to function properly by using an ip it grabbed (even with error during process), but now this function will cause an error during the init instead - is this something worth considering?

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,218 @@
...
+			"node IP %s is not in any of the remote network CIDR blocks: %s. "+
+				"See https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-troubleshooting.html or use --skip node-ip-validation",
+			ipAddr, nodeNetworkCidrs)
+	}
+	return nil
+}
+
+func (hnp *HybridNodeProvider) ValidateIP() error {
+	if hnp.cluster == nil {
+		hnp.Logger().Info("Node IP validation skipped")
```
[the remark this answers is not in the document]
achen1210: yeah makes sense

internal/node/hybrid/ip_validator.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil, err
+	}
+
+	var ipAddr net.IP
+
+	nodeIPSpecified := nodeIP != nil && nodeIP.To4() != nil && !nodeIP.IsUnspecified()
+
+	if nodeIPSpecified {
+		ipAddr = nodeIP
+	} else {
```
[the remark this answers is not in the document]
achen1210: yeah right now only the ipAddr = nodeIP branch is tested (since this is when ip is set via --node-ip kubelet flag), the intent was to use integration tests for the other 2 branches (dns & resolvebindaddress) however those would require mocking the describecluster call for the remotenetworkconfig hence the change to moto.

for unit testing I think it would require adding an interface to do the mocking, i.e. `type IPLookup interface {
    LookupIP(host string) ([]net.IP, error)
}` and then make a real implementation and a mocked implementation, although I'm not sure if its worth it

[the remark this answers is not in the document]
achen1210: added unit tests with the interface for mocking

internal/node/hybrid/ip_validator.go
achen1210: node-ip and hostname-override can only be set via flags under my understanding; there's no option to set it through a kubelet config field (https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/)

internal/node/hybrid/hybrid.go
```diff
@@ -1,28 +1,39 @@
...
 
 type NodeProviderOpt func(*HybridNodeProvider)
 
-func NewHybridNodeProvider(nodeConfig *api.NodeConfig, logger *zap.Logger, opts ...NodeProviderOpt) (nodeprovider.NodeProvider, error) {
+func NewHybridNodeProvider(nodeConfig *api.NodeConfig, skipPhases []string, logger *zap.Logger, opts ...NodeProviderOpt) (nodeprovider.NodeProvider, error) {
 	np := &HybridNodeProvider{
 		nodeConfig: nodeConfig,
 		logger:     logger,
+		skipPhases: skipPhases,
+		network:    &defaultKubeletNetwork{},
```
g-gaston, line 36: nit:
```suggestion
		network:    defaultKubeletNetwork{},
```

g-gaston (approved):

PR #359, opened 2025-02-20
Enable cluster logging for test clusters

abhinavmpandey08:
*Description of changes:*
Enable cluster logging for test clusters


g-gaston (approved):
Are the logs persisted even after the cluster is deleted?
If so, for how long?

abhinavmpandey08:
>Are the logs persisted even after the cluster is deleted?
If so, for how long?

Yes the logs are persisted even after the cluster is deleted. Looks like the retention policy is so the logs never expire.

abhinavmpandey08:
https://docs.aws.amazon.com/eks/latest/best-practices/cost-opt-observability.html#_eks_data_plane

PR #363, opened 2025-02-21
Implement routable pod CIDRs in e2e tests

g-gaston:
## Description of changes
We now create an individual route per hybrid node on the subnet where the hybrid nodes live for the pod CIDR assigned to the node using the instance as the next hop.

In order for this work, it required swapping the peered VPC for a TGW. Not because the individual routes couldn't be added, but because the VPC networking would refuse to forward a packet through the peered connection with a dst IP that doesn't belong to the VPC'S CIDR blocks on the other side. This is a problem because the hybrid pod CIDR is not part of the second VPC. And it can't be if we want to be able to add the previously mentioned per-node routes.

This enables CP to pod communication or east-west pod to pod communication.

*Testing (if applicable):*
Running this in my stack.


test/e2e/kubernetes/node.go
g-gaston, line 1: most of these are just moved from `test/e2e/kubernetes/kubernetes.go`
the only thing new are the last 2 methods

test/e2e/cni/podcidr.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func ciliumNodePodCIDR(ctx context.Context, k8s dynamic.Interface, node *corev1.Node) ([]string, error) {
+	ciliumNodeGVR := schema.GroupVersionResource{
+		Group:    "cilium.io",
+		Version:  "v2",
+		Resource: "ciliumnodes",
+	}
+
+	obj, err := k8s.Resource(ciliumNodeGVR).Get(ctx, node.Name, metav1.GetOptions{})
```
jaxesn, line 46: we should probably retry these k8s requests.

test/e2e/cni/podcidr.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	if len(podCIDRs) == 0 {
+		return nil, fmt.Errorf("no calico IPAMBlock found for node %s", node.Name)
+	}
+
+	return podCIDRs, nil
+}
+
+type ipamBlock struct {
```
jaxesn, line 97: does this only apply to calico? do you want to name as such?

test/e2e/ec2/instance.go
```diff
@@ -145,6 +145,20 @@ func (e *InstanceConfig) Create(ctx context.Context, ec2Client *ec2.Client, ssmC
 	}, nil
 }
 
+// DisableSourceDestCheck disables the source/destination check for the given instance.
+func DisableSourceDestCheck(ctx context.Context, ec2Client *ec2.Client, instanceID string) error {
```
jaxesn, line 149: so by default an ec2 instance blocks the packet if the src/dest is outside the vpc(?) ?

test/e2e/peered/node.go
```diff
@@ -224,6 +234,12 @@ func (c *NodeCleanup) Cleanup(ctx context.Context, instance ec2.Instance) error
 		c.Logger.Info("Skipping EC2 Instance deletion", "instanceID", instance.ID)
 		return nil
 	}
+
+	c.Logger.Info("Deleting routes for EC2 Instance", "instanceID", instance.ID, "subnetID", c.Cluster.SubnetID)
+	if err := ec2.DeleteRoutesForInstance(ctx, c.EC2, c.Cluster.SubnetID, instance.ID); err != nil {
```
jaxesn, line 239: is there any issue to other tests, potentially, if this fails to cleanup or just in general?

jaxesn (commented):
probably going to need new perms. otherwise, just a couple questions mainly

test/e2e/ec2/instance.go
g-gaston: yeah

> You can enable or disable source/destination checks, which ensure that the instance is either the source or the destination of any traffic that it receives. Source/destination checks are enabled by default. You must disable source/destination checks if the instance runs services such as network address translation, routing, or firewalls.

test/e2e/peered/node.go
g-gaston: maybe

if the route is not deleted and we try to create another route for the same cidr, it will fail. So it's possible for:
1. a node to be deleted
2. its route not be deleted
3. then another node is created
4. cilium happens to assign the same pod cidr to this new node that it had assigned to the previous one
5. when the test gets to create the vpc route, it will be duplicated and fail.

The alternative is to detect the duplicated route error in the test and delete it first.

I opted not to do it because that could cause another running test to fail if it happens that the other node hasn't been deleted yet. That code could be masking a different type of issue. Moreover, the test error would be very difficult to debug. 

With the current code, the error will be obvious. If this starts being a problem, we can reconsider it.

test/e2e/cni/podcidr.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if !found {
+		return nil, fmt.Errorf("podCIDRs not found in CiliumNode %s", node.Name)
+	}
+	return podCIDRs, nil
+}
+
+func calicoNodePodCIDR(ctx context.Context, k8s dynamic.Interface, node *corev1.Node) ([]string, error) {
+	var podCIDRs []string
+	err := retry.OnError(retry.DefaultBackoff, func(err error) bool {
```
g-gaston, line 62: why do we need this retry if we are already using RetryList?

test/e2e/peered/node.go
```diff
@@ -151,13 +151,37 @@ func (c NodeCreate) Create(ctx context.Context, spec *NodeSpec) (PeerdNode, erro
...
+func (c *NodeCreate) WaitForNodeReady(ctx context.Context, verifyNode *kubernetes.VerifyNode, peeredNetwork *Network, peeredNode *PeerdNode) error {
+	_, err := verifyNode.WaitForNodeReady(ctx)
+	if err != nil {
+		return fmt.Errorf("waiting node object in kubernetes to be ready: %w", err)
+	}
+
+	// Configure PodCIDR for EC2 route table
+	if err := peeredNetwork.CreateRoutesForNode(ctx, peeredNode); err != nil {
+		return fmt.Errorf("creating routes for peered node: %w", err)
+	}
```
g-gaston, line 180: I don't think this belong here. This method waits for a node to be ready. Why would it have a side effect of creating routes?

test/e2e/peered/node.go
```diff
@@ -151,13 +151,37 @@ func (c NodeCreate) Create(ctx context.Context, spec *NodeSpec) (PeerdNode, erro
...
+
 	c.Logger.Info("A Hybrid EC2 instace is created", "instanceID", instance.ID)
 	return PeerdNode{
 		Instance: instance,
 		Name:     spec.NodeName,
 	}, nil
 }
 
+func (c *NodeCreate) WaitForNodeReady(ctx context.Context, verifyNode *kubernetes.VerifyNode, peeredNetwork *Network, peeredNode *PeerdNode) error {
```
g-gaston, line 171: I would prefer if wait for node ready didn't depend on the peered package
that should be a pure k8s operation

test/e2e/peered/node.go
```diff
@@ -292,12 +322,25 @@ func (c *NodeCleanup) Cleanup(ctx context.Context, node PeerdNode) error {
...
 
+func (c *NodeCleanup) Reset(ctx context.Context, cleanNode *nodeadm.CleanNode, node *PeerdNode) error {
+	if err := cleanNode.Run(ctx); err != nil {
+		return err
+	}
+
+	// Delete routes for EC2 Instance
+	if err := ec2.DeleteRoutesForInstance(ctx, c.EC2, c.Cluster.SubnetID, node.Instance.ID); err != nil {
+		return fmt.Errorf("deleting routes for EC2 Instance: %w", err)
+	}
```
g-gaston, line 333: How does this solve the problem?
Cleanup will run uninstall and then delete the node object. This deletes the routes after the node object is deleted, which is in fact the root cause of the race condition. We need to delete the routes before the node object, so cilium/calico don't assigned the same slice of pod cidr to a different node before the routes are deleted.

g-gaston: Once thing we could do is to inject a "NodeInfrastructureCleaner" in the cleanup object. That could be implemented by our peered node and run the routes cleanup. We would run that before actually deleting the node object from the cleanup.

We already have that pattern with the `UninstallVerifier`

test/e2e/cni/podcidr.go
junpengdev: RetryList would not retry if there is no error calling kubernetes APIs. What happens here is when this gets called, ipamblocks CRD for the node either doesn't exist or doesn't have pod cidr yet.

test/e2e/peered/node.go
junpengdev: Maybe we can rename it to something else, like `waitForPeeredNodeReady`? With current network setup, a hybrid is ready only when 1) node is ready in kubernetes, and 2) route is configured in ec2 instance to route traffic to pod.

test/e2e/peered/node.go
junpengdev: Sure, I will define `NodeInfrastructureCleaner` interface and add a field to cleanup object.

test/e2e/peered/node.go
junpengdev: I can rename the method to something more meaningful.

test/e2e/nodeadm/uninstall.go
```diff
@@ -13,21 +13,32 @@ import (
...
-	Logger              logr.Logger
+	K8s                   clientgo.Interface
+	RemoteCommandRunner   commands.RemoteCommandRunner
+	Verifier              UninstallVerifier
+	InfrastructureCleaner NodeInfrastructureCleaner
+	Logger                logr.Logger
 
 	NodeName string
 	NodeIP   string
+	Instance string
```
g-gaston, line 24: this ties the cleanup to having an instance
this is supposed to be a generic "nodeadm clean node", hence why we are introducing a new interface
But the nodes doesn't need to be backed by an instance
can we avoid this?
can't we just pass the node name?

test/e2e/peered/node.go
```diff
@@ -294,12 +310,21 @@ func (c *NodeCleanup) Cleanup(ctx context.Context, node PeerdNode) error {
...
+}
+
 func (c Node) S3LogsURL(instanceName string) string {
-	return fmt.Sprintf("https://%s.console.aws.amazon.com/s3/buckets/%s?prefix=%s/", c.Cluster.Region, c.LogsBucket, c.logsPrefix(instanceName))
+	return fmt.Sprintf("https://%s.console.aws.amazon.com/s3/buckets/%s?prefix=%s/", c.NodeCreate.Cluster.Region, c.LogsBucket, c.logsPrefix(instanceName))
 }
 
 func (c NodeCleanup) logsPrefix(instanceName string) string {
-	return fmt.Sprintf("%s/%s/%s", constants.TestS3LogsFolder, c.ClusterName, instanceName)
+	return fmt.Sprintf("logs/%s/%s", c.Cluster.Name, instanceName)
```
g-gaston, line 327: why is this changing?

junpengdev: This is from your initial commits. Can you recall why?

test/e2e/nodeadm/uninstall.go
junpengdev: changed it in this commit: https://github.com/aws/eks-hybrid/pull/363/commits/d79e6bc5f1c46ed5d225979c6f1c1529ec794d2d

let me know what you think.

test/e2e/peered/node.go
jaxesn: this could be a merge conflict issue, that looks like a change i made in a pr recently. double check whats in main now

test/e2e/cni/podcidr.go
jaxesn: lets add a comment that explains that so its clear we are retrying here because we know the object will take time to exist vs just handling timeout issues.

test/e2e/ec2/instance.go
```diff
@@ -145,6 +145,20 @@ func (e *InstanceConfig) Create(ctx context.Context, ec2Client *ec2.Client, ssmC
...
+	_, err := ec2Client.ModifyInstanceAttribute(ctx, &ec2.ModifyInstanceAttributeInput{
```
jaxesn, line 150: did we add a new perm to cover this api call?

test/e2e/peered/node.go
junpengdev: I didn't see it in the merge conflict check in github UI. Here is what we have in `main`.

```
func (c NodeCleanup) logsPrefix(instanceName string) string {
	return fmt.Sprintf("%s/%s/%s", constants.TestS3LogsFolder, c.ClusterName, instanceName)
}
```

test/e2e/peered/network.go
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Errorf("describing route tables: %w", err)
+	}
+
+	if len(resp.RouteTables) != 1 {
+		return fmt.Errorf("expected to find one route table for subnet %s, found %d", n.Cluster.SubnetID, len(resp.RouteTables))
+	}
+
+	for _, cidr := range cidrs {
+		n.Logger.Info("Adding route for CIDR", "cidr", cidr, "instanceID", instance.ID)
+		err := ec2.CreateRouteForCIDRToInstance(ctx, n.EC2, *resp.RouteTables[0].RouteTableId, cidr, instance.ID)
```
jaxesn, line 66: same here, was a new perm we needed?

test/e2e/ec2/instance.go
junpengdev: Yes. `ec2:ModifyInstanceAttribute` was added in #461

test/e2e/suite/nodeadm/nodeadm_test.go
```diff
@@ -111,6 +112,7 @@ var _ = Describe("Hybrid Nodes", func() {
 					testNode.It("re-joins the cluster after reboot", func() {
 						Expect(testNode.WaitForNodeReady(ctx)).Error().To(Succeed(), "node should have re-joined, there must be a problem with uninstall")
 					})
+					Expect(testNode.PeeredNetwork.CreateRoutesForNode(ctx, n)).Should(Succeed(), "EC2 route to pod CIDR should have been created successfully")
```
jaxesn, line 115: do we need this here since we also have it test_node?

what about the conformance test, do we need similar changes there?

junpengdev: Yes, we need it here. The route will be deleted when node object is deleted from kubernetes. When rebooting ec2 instance, cilium can potentially assign a different pod cidr to the instance. 
We had it in test_node but not in `waitForNodeReady` function. Thus we need to call it again in `nodeadm_test.go`

junpengdev: Just skimmed the conformance tests. I don't see we have reboot scenario thus I don't think we need it there. Let me know if I miss anything.

jaxesn: can we jsut put it in testNode.WaitForNodeReady so its only in the test-node code?

test/e2e/suite/nodeadm/nodeadm_test.go
```diff
@@ -154,7 +156,10 @@ var _ = Describe("Hybrid Nodes", func() {
 						test.Logger.Info("Skipping nodeadm uninstall from the hybrid node...")
 						return
 					}
-					Expect(test.NewCleanNode(provider, testNode.PeerdNode().Name, testNode.PeerdNode().Instance.IP).Run(ctx)).To(
+
+					n := testNode.PeerdNode()
+					cleanNode := test.NewCleanNode(provider, testNode.PeeredNode.CreateNodeInfrastrastructureCleaner(n.Instance.ID), n.Name, n.Instance.IP)
```
jaxesn, line 161: is the cleanode the only place we run this infracleaner?  in the case of conformance we dont call clean node, we just do the normal cleanup in NodeCleanup.Cleanup.

test/e2e/peered/node.go
jaxesn: yeah the issue this PR changed where it is getting the cluster.name, so it conflicted at somepoint when me adding the constant. make it use the constant instead of "logs"

test/e2e/suite/nodeadm/nodeadm_test.go
jaxesn: yeah i get it now, what you have is good. this is the only case we are testing the reboot/reset.

test/e2e/peered/node.go
g-gaston: let's just revert this line pls

junpengdev (approved):

junpengdev (approved):

PR #364, opened 2025-02-21
adds more cleanup to e2e-test cleanup subcommand

jaxesn:
*Issue #, if available:*

*Description of changes:*

There are multiple levels of cleanup in our e2e test.  The normal testing flow goes like:

- e2e-test setup... to create the Infra CFN Stack and the EKS Cluster, along with a handful of other resources
- ginkgo run... runs a single (or multiple tests in parallel), either way at the beginning on such a run, the Credential CFN Stack is created, along other resources such as the  access entries
- when a test completes, it cleans up the EC2 instance associated with that test
- when all the tests complete, the cred cfn stack is deleted, along with the access entry and other resources
- in ci, e2e-test cleanup is than ran to delete the eks cluster and then the infra stack
- at the end of the ci run (and beginning), we also run the e2e_cleanup.sh script to catch any potential orphaned resources

Currently, ssm activations and managed instances are the only (there could be another i am forgetting) resources that are not cleaned up by the e2e-test/gingko assuming everything completes successfully.

This PR adds activation/managed instance cleanup to tests themselves which would mean if everything works correctly, the bash script will not do anything.

We should think of this current bash script as a failsafe to clean up anything our test cleanup and/or cloudformation leaves around.  We want to keep this process.  That said, this PR also starts the process of moving this logic into go to better fit the rest of the code base and reuse as much as possible.  There are 3 use cases for this new cleanup flow, which match more or less how we use the bash script today:
- cleanup with a specific cluster name. this is run at the end of a CI build or Canary run and it deletes everything associated with that cluster regardless of its age
- cleanup with a cluster name prefix. this is run at the beginning of a canary run for a specific kube version, ex: nodeadm-e2e-1-31-, this will delete all clusters and resources with that prefix regardless of their age
- cleanup with no cluster prefix/name, this is run currently at the beginning of a ci run and intended to delete all clusters a day old

This PR models these use cases with flags on the e2e-test cleanup subcommand.

*Testing (if applicable):*


test/e2e/cleanup/eks.go
```diff
@@ -0,0 +1,84 @@
...
+	resource := ResourceWithTags{
+		ID:           *cluster.Name,
+		CreationTime: aws.ToTime(cluster.CreatedAt),
+		Tags:         tags,
+	}
+	return shouldDeleteResource(resource, input)
+}
+
+// arn format: arn:aws:eks:us-west-2:<account-id>:cluster/nodeadm-e2e-tests-1-31
+func extractClusterName(stackARN string) string {
```
g-gaston, line 74: nit
```suggestion
func extractClusterName(clusterARN string) string {
```

test/e2e/cleanup/eks.go
```diff
@@ -0,0 +1,84 @@
...
+	Logger          logr.Logger
+}
+
+func (c *EKSClusterCleanup) ListEKSClusters(ctx context.Context, input FilterInput) ([]string, error) {
+	tagger := &ResourceTagger{
+		ResourceTagging: c.ResourceTagging,
+		ClusterName:     input.ClusterName,
+	}
+
+	stackARNs, err := tagger.GetTaggedResources(ctx, "eks:cluster")
```
g-gaston, line 29:
```suggestion
	clusterARNs, err := tagger.GetTaggedResources(ctx, "eks:cluster")
```

test/e2e/cleanup/iam.go
```diff
@@ -0,0 +1,176 @@
...
+	// but we only delete those with the e2e cluster tag
+	paginator := iam.NewListRolesPaginator(c.IAM, &iam.ListRolesInput{})
+
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("listing IAM roles: %w", err)
+		}
+		for _, role := range page.Roles {
+			if !strings.HasPrefix(*role.RoleName, constants.TestCredentialsStackNamePrefix) {
```
g-gaston, line 36: should use the TestCredentialsStackNamePrefix in the form of `/prefix/whatever` when creating the roles so we can use the prefix path argument for list roles?

test/e2e/cleanup/iam.go
```diff
@@ -0,0 +1,176 @@
...
+	return shouldDeleteResource(resource, input)
+}
+
+func (c *IAMCleaner) DeleteRole(ctx context.Context, roleName string) error {
+	instanceProfiles, err := c.IAM.ListInstanceProfilesForRole(ctx, &iam.ListInstanceProfilesForRoleInput{
+		RoleName: aws.String(roleName),
+	})
+	if err != nil && errors.IsType(err, &types.NoSuchEntityException{}) {
+		c.Logger.Info("IAM instance profiles already deleted", "roleName", roleName)
+		return nil
```
g-gaston, line 81: don't we just want to continue and skip the instance profile part but still delete the role?

test/e2e/cluster/delete.go
```diff
@@ -48,20 +66,224 @@ func NewDelete(aws aws.Config, logger logr.Logger, endpoint string) Delete {
...
+
+	if err := c.cleanupCredentialStacks(ctx, filterInput); err != nil {
+		return fmt.Errorf("cleaning up credential stacks: %w", err)
+	}
+
+	if err := c.cleanupEKSClusters(ctx, filterInput); err != nil {
+		return fmt.Errorf("cleaning up EKS clusters: %w", err)
+	}
+
+	if err := c.cleanupArchitectureStacks(ctx, filterInput); err != nil {
```
g-gaston, line 98: i think for the credentials stack we already have a fail-safe cleanup of individual resources. should we also have it for this one? like vpc, subnets, route tables, vpc peering...

test/e2e/cleanup/iam.go
```diff
@@ -0,0 +1,207 @@
...
+		iamClient: iamClient,
+		logger:    logger,
+	}
+}
+
+func (c *IAMCleaner) ListRoles(ctx context.Context, filterInput FilterInput) ([]string, error) {
+	var roles []string
+
+	paginator := iam.NewListRolesPaginator(c.iamClient, &iam.ListRolesInput{
+		PathPrefix: aws.String(constants.TestRolePathPrefix),
```
g-gaston, line 32: don't we need to maintain the old cleanup for roles for a bit? at least for a few rounds until we are sure everything has been deleted?

jaxesn: I'm going to keep the bash script around for a bit to cover this case.

cmd/e2e-test/cleanup/cleanup.go
```diff
@@ -63,9 +65,24 @@ func (s *Command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 
 	delete := cluster.NewDelete(aws, logger, deleteCluster.Endpoint)
 
+	// run the standard cluster cleanup as well as the sweeper fallback
 	logger.Info("Cleaning up E2E cluster resources...")
-	if err = delete.Run(ctx, deleteCluster); err != nil {
-		return fmt.Errorf("error cleaning up e2e resources: %w", err)
+
+	if deleteErr := delete.Run(ctx, deleteCluster); deleteErr != nil {
```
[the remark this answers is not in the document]
jaxesn: Yeah we can. As far as the command itself, that probably makes sense. From test run perspective, CI or canary, I'd still like to run the sweeper like we do now with the cleanup subcommand and the bash script. 

I put both here to just sorta hide/wrap that. But I have no problem adding it as a separate call in the run-e2e/canary flow.

test/e2e/cluster/cfn-templates/setup-cfn.yaml
```diff
@@ -80,7 +84,9 @@ Resources:
           Value: !Sub ${ClusterName}-cluster-vpc
         - Key: !Ref TestClusterTagKey
           Value: !Ref ClusterName
-  
+        - Key: CreationTime
```
g-gaston, line 87: can we make this key a param so we can use the same constant we use in the go code when listing?

test/e2e/cluster/stack.go
```diff
@@ -121,6 +121,10 @@ func (s *stack) deploy(ctx context.Context, test TestResources) (*resourcesStack
 			ParameterKey:   aws.String("RolePathPrefix"),
 			ParameterValue: aws.String(constants.TestRolePathPrefix),
 		},
+		{
+			ParameterKey:   aws.String("CreationTime"),
+			ParameterValue: aws.String(time.Now().Format(time.RFC3339)),
```
g-gaston, line 126: this probably updates the tag value every time you the `setup` command, right? which means all resources with the tag get updated

g-gaston (approved):

PR #369, opened 2025-02-28
Validate cni ports before running init

vignesh-goutham:
*Description of changes:*
Validate if firewall is enabled and running on the host, if it is, make sure either of the support CNIs (cilium/calico) Vxlan ports have been opened. If not alert the user to allow the ports for proper cluster operations. Users can override this check with a `--skip cni-validation` flag to the init command.

*Testing (if applicable):*
Tested on Ubuntu for ufw and Redhat for firewalld


cmd/nodeadm/init/init.go
```diff
@@ -93,3 +108,30 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
...
+	if err != nil {
+		return err
+	}
+	if !enabled {
+		return nil
+	}
+	if err := firewallManager.FlushRules(); err != nil {
+		return err
+	}
+	ciliumVxlanPortOpen, err := firewallManager.IsPortOpen(ciliumVxLanPort, vxLanProtocol)
```
junpengdev, line 124: If port is not open, then can we configure it with `nodeadm`?

cmd/nodeadm/init/init.go
```diff
@@ -93,3 +108,30 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
...
+	if err != nil {
+		return err
+	}
+	calicoVxlanPortOpen, err := firewallManager.IsPortOpen(calicoVxLanPort, vxLanProtocol)
+	if err != nil {
+		return err
+	}
+
+	if !ciliumVxlanPortOpen && !calicoVxlanPortOpen {
```
junpengdev, line 133: If customer has calico ports open but uses cilium, then the validation can pass but the CNI may still not work as expected.

internal/firewall/ufw.go
```diff
@@ -63,3 +79,42 @@ func (ufw *UncomplicatedFireWall) FlushRules() error {
 	// UFW activates the rules the moment its added, there is no need to flush them out to disk explicitly
 	return nil
 }
+
+func (ufw *UncomplicatedFireWall) IsPortOpen(port, protocol string) (bool, error) {
+	if len(ufw.rules) == 0 {
```
junpengdev, line 84: curious why we need to check length here. Can we just refresh firewall rules directly?

junpengdev (changes requested):

internal/firewall/ufw.go
vignesh-goutham: Sure we can, just a minute operational efficiency. If its already refreshed I dont want to run it again.

cmd/nodeadm/init/init.go
vignesh-goutham: Yes, at the time of bootstrapping we do not know with certainty what CNI the cluster is going to use. We validate best effort to make user user has a good chance of success.

cmd/nodeadm/init/init.go
```diff
@@ -93,3 +108,30 @@ func (c *initCmd) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
...
+		return nil
+	}
+	if err := firewallManager.FlushRules(); err != nil {
+		return err
+	}
+	ciliumVxlanPortOpen, err := firewallManager.IsPortOpen(ciliumVxLanPort, vxLanProtocol)
+	if err != nil {
+		return err
+	}
+	calicoVxlanPortOpen, err := firewallManager.IsPortOpen(calicoVxLanPort, vxLanProtocol)
```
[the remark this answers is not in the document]
vignesh-goutham: We are actually validating only if either one port is open.

test/e2e/os/testdata/rhel/8/cloud-init.txt
```diff
@@ -25,6 +25,6 @@ write_files:
 {{- end }}
 
 runcmd:
-  - /tmp/nodeadm-init.sh "{{ .NodeadmUrl }}" "{{ .KubernetesVersion }}" "{{ .Provider }}" "--containerd-source docker"
+  - /tmp/nodeadm-init.sh "{{ .NodeadmUrl }}" "{{ .KubernetesVersion }}" "{{ .Provider }}" "--containerd-source docker" "--skip cni-validation"
```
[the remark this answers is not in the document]
vignesh-goutham: Actually firewalld isnt running on the current amis we pick up for testing. I put this in, just in case. I've removed them.

junpengdev (approved):

PR #370, opened 2025-02-28
adds retry for ec2 instance creation

jaxesn:
*Issue #, if available:*

*Description of changes:*

We have noticed a fair amount of ec2 instances that for some reason do not appear to get far enough into the boot/cloud-init process to join as eks nodes.  This PRs uses the instancestatus to check for impaired instances.  Impaired is being defined as the reachability check has failed.  If the node fails to join as normal, we check if its impaired and retry if so.

To handle this retry we can not use the default ginkgo retry mechanism because we would need to break our test up into multiple Its.  This was tried in https://github.com/aws/eks-hybrid/pull/317.  This would work and probably be simpler than this PR, however, once we break them into multiple It blocks, the test output in CI will be all munged together.  I do not think that would be a good change.

This introduces a FlakeAttempt "It" so we can wrap the code which creates the ec2 instance in a block that will retry if we trigger a retryablexpect.

Additional Notes

I tested creating instances and tracking when the instance status changes and it does not seem to be related to cloud-init timings and/or error codes. The node checks (as well as the healthy status) goes green a minute or more after the cloud-init finishes.  I think this also gives us some confidence that if the cloud-init script is running and fails, that would not affect the ec2 healthchecks. So we can rely on this reachability check to determine if its ok to delete and retry the node.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/ec2/instance.go
+++ b/test/e2e/ec2/instance.go
@@ -163,6 +163,28 @@ func WaitForEC2InstanceRunning(ctx context.Context, ec2Client *ec2.Client, insta
 	return waiter.Wait(ctx, &ec2.DescribeInstancesInput{InstanceIds: []string{instanceID}}, nodeRunningTimeout)
 }
 
+func IsEC2InstanceImpaired(ctx context.Context, ec2Client *ec2.Client, instanceID string) (bool, error) {
+	describeStatusOutput, err := ec2Client.DescribeInstanceStatus(ctx, &ec2.DescribeInstanceStatusInput{
+		InstanceIds:         []string{instanceID},
+		IncludeAllInstances: aws.Bool(true),
+	})
+	if err != nil {
+		return false, fmt.Errorf("describing instance status %s: %w", instanceID, err)
+	}
+	isImpaired := false
+	for _, status := range describeStatusOutput.InstanceStatuses[0].SystemStatus.Details {
+		if status.Name == "reachability" {
+			isImpaired = true
+		}
+	}
+	for _, status := range describeStatusOutput.InstanceStatuses[0].InstanceStatus.Details {
+		if status.Name == "reachability" {
+			isImpaired = true
+		}
+	}
+	return isImpaired, nil
+}
+
 func LogEC2InstanceDescribe(ctx context.Context, ec2Client *ec2.Client, instanceID string, logger logr.Logger) error {
 	instances, ec2Err := ec2Client.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
 		InstanceIds: []string{instanceID},
--- a/test/e2e/suite/flake_attempt.go
+++ b/test/e2e/suite/flake_attempt.go
@@ -1,0 +1,99 @@
+package suite
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
+)
+
+type FlakeAttempt struct {
+	logger logr.Logger
+}
+type FlakeRun struct {
+	DeferCleanup    func(func(context.Context), ...interface{})
+	RetryableExpect func(actual interface{}, extra ...interface{}) Assertion
+}
+
+type IgnorablePanic struct{}
+
+func (IgnorablePanic) GinkgoRecoverShouldIgnoreThisPanic() {}
+
+func (f *FlakeAttempt) It(ctx context.Context, description string, flakeAttempts int, body func(ctx context.Context, flakeRun FlakeRun)) {
+	var lastErr error
+	for attempt := range flakeAttempts {
+		// register globally as well in case the test fails for any reason
+		// including being cancelled while this block is executing
+		// track if the cleanup runs and do not run it again if it does
+		cleanups := []func(context.Context){}
+		deferCleanup := func(cleanup func(context.Context), args ...interface{}) {
+			ran := false
+			onced := func(ctx context.Context) {
+				if ran {
+					f.logger.Info(fmt.Sprintf("Cleanup already ran for flake attempt %d, skipping", attempt+1))
+					return
+				}
+				ran = true
+				cleanup(ctx)
+			}
+			cleanups = append(cleanups, onced)
+			DeferCleanup(onced, args)
+		}
+
+		testGoMega := gomega.NewGomega(func(message string, callerSkip ...int) {
+			lastErr = fmt.Errorf("%s", message)
+			panic(IgnorablePanic{})
+		})
+		flakeRun := FlakeRun{
+			DeferCleanup:    deferCleanup,
+			RetryableExpect: testGoMega.Expect,
+		}
+
+		c := make(chan error)
+		go func() {
+			defer func() {
+				var err error
+				// if this was a ignorable panic, we send nil
+				// to indicate that we should retry
+				// otherwise we send the error to end execution
+				e := recover()
+				if lastErr == nil && e != nil {
+					err = e.(error)
+				}
+				c <- err
+			}()
+			By(description)
+			body(ctx, flakeRun)
+		}()
+
+		var panicableError error
+		Eventually(c).WithContext(ctx).Should(Receive(&panicableError))
+		Expect(panicableError).NotTo(HaveOccurred())
+
+		if lastErr == nil {
+			if attempt > 1 {
+				f.logger.Info(fmt.Sprintf("Succeeded on attempt %d after previous failures", attempt+1))
+			}
+			break
+		}
+
+		if attempt == flakeAttempts-1 {
+			break
+		}
+
+		// omly run cleanups if about to retry, otherwise let gingko handle it
+		for _, f := range slices.Backward(cleanups) {
+			f(ctx)
+		}
+
+		f.logger.Error(lastErr, fmt.Sprintf("Failed attempt %d/%d", attempt+1, flakeAttempts))
+		time.Sleep(time.Second * time.Duration(attempt))
+
+	}
+	Expect(lastErr).To(BeNil(), "should have succeeded")
+}
--- a/test/e2e/suite/nodeadm_test.go
+++ b/test/e2e/suite/nodeadm_test.go
@@ -240,40 +240,62 @@ var _ = Describe("Hybrid Nodes", func() {
 							if test.overrideNodeK8sVersion != "" {
 								k8sVersion = suite.TestConfig.NodeK8sVersion
 							}
-
 							peeredNode := test.newPeeredNode()
-							instance, err := peeredNode.Create(ctx, &peered.NodeSpec{
-								InstanceName:   instanceName,
-								NodeK8sVersion: k8sVersion,
-								NodeNamePrefix: "simpleflow",
-								OS:             nodeOS,
-								Provider:       provider,
-							})
-							Expect(err).NotTo(HaveOccurred(), "EC2 Instance should have been created successfully")
-							DeferCleanup(func(ctx context.Context) {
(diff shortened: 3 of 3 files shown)


test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		c := make(chan error)
+		go func() {
+			defer func() {
+				var err error
+				// if this was a ignorable panic, we send nil
+				// to indicate that we should retry
+				// otherwise we send the error to end execution
+				e := recover()
```
jaxesn, line 64: this is very similar to the ginkgo provided GinkgoRecover. The problem with using it directly is we have no idea, and i could not find a way to determine it from ginkgo global state, if an error has happened and we want to end the test.  This approach allows us to get the error and recover from the panic, but if the error is our retryable error, we "ignore" it.  In the case that the error is not our retryable, then it will fall down into the Eventually and fail the overall test.

test/e2e/suite/nodeadm_test.go
```diff
@@ -240,40 +240,62 @@ var _ = Describe("Hybrid Nodes", func() {
...
+									_, err := verifyNode.WaitForNodeReady(ctx)
+									if err != nil {
+										isImpaired, oErr := ec2.IsEC2InstanceImpaired(ctx, test.ec2Client, instance.ID)
+										if oErr != nil {
+											Expect(err).NotTo(HaveOccurred(), "should describe instance status")
+										}
+										expect := Expect
+										if isImpaired {
+											expect = flakeRun.RetryableExpect
+										}
```
g-gaston, line 290: maybe comment explaining the rationale here? like why retry if impaired. i think th code is very easy to read but the reasons to write this code maybe not

test/e2e/ec2/instance.go
```diff
@@ -163,6 +163,28 @@ func WaitForEC2InstanceRunning(ctx context.Context, ec2Client *ec2.Client, insta
...
+		InstanceIds:         []string{instanceID},
+		IncludeAllInstances: aws.Bool(true),
+	})
+	if err != nil {
+		return false, fmt.Errorf("describing instance status %s: %w", instanceID, err)
+	}
+	isImpaired := false
+	for _, status := range describeStatusOutput.InstanceStatuses[0].SystemStatus.Details {
+		if status.Name == "reachability" {
+			isImpaired = true
```
g-gaston, line 177: nit can we just return
```suggestion
			return true
```

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+	"sync"
+	"time"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
+)
+
+type FlakeAttempt struct {
```
g-gaston, line 16: doc comment explaining what this is for?

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+	logger logr.Logger
```
g-gaston, line 17: if we are gonna make the struct exported, let's make the field exported as well

test/e2e/suite/nodeadm_test.go
```diff
@@ -240,40 +240,62 @@ var _ = Describe("Hybrid Nodes", func() {
...
-										". You can access the collected node logs at: %s", peeredNode.S3LogsURL(instance.Name),
-								)
+							var verifyNode *kubernetes.VerifyNode
+							var serialOutput peered.ItBlockCloser
+							var instance ec2.Instance
+
+							flakeAttempt := &FlakeAttempt{
+								logger: test.logger,
+							}
+							flakeAttempt.It(ctx, "Creating a node", 3, func(ctx context.Context, flakeRun FlakeRun) {
```
g-gaston, line 252: nit
```suggestion
							flakeAttempt.It(ctx, "Creates a node", 3, func(ctx context.Context, flakeRun FlakeRun) {
```

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+				}
+				ran = true
+				cleanup(ctx)
+			}
+			cleanups = append(cleanups, onced)
+			DeferCleanup(onced, args)
+		}
+
+		testGoMega := gomega.NewGomega(func(message string, callerSkip ...int) {
+			lastErr = fmt.Errorf("%s", message)
```
g-gaston, line 50: nit
```suggestion
			lastErr = errors.New(message)
```

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+				// to indicate that we should retry
+				// otherwise we store the panicable error to trigger
+				// the test to fail
+				e := recover()
+				if lastErr == nil && e != nil {
+					panicableError = e.(error)
+				}
+			}()
+			By(description)
+			body(ctx, flakeRun)
```
g-gaston, line 74: nit: create copy of context?

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+		go func() {
+			defer wg.Done()
+			defer func() {
+				// if this was a ignorable panic, we ignore it
+				// to indicate that we should retry
+				// otherwise we store the panicable error to trigger
+				// the test to fail
+				e := recover()
+				if lastErr == nil && e != nil {
+					panicableError = e.(error)
```
g-gaston, line 70: probably should check if this is an error? i know it should be, but if something else panic with something but an error, i think this will panic

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+	logger logr.Logger
+}
+type FlakeRun struct {
+	DeferCleanup    func(func(context.Context), ...interface{})
+	RetryableExpect func(actual interface{}, extra ...interface{}) Assertion
+}
+
+type IgnorablePanic struct{}
+
+func (IgnorablePanic) GinkgoRecoverShouldIgnoreThisPanic() {}
```
g-gaston, line 26: do we need this anymore? it sees like when we want to panic with a retryable error, it doesn't matter what we panic with?

test/e2e/suite/flake_attempt.go
```diff
@@ -0,0 +1,101 @@
...
+		}()
+
+		wg.Wait()
+		Expect(panicableError).NotTo(HaveOccurred())
+
+		if lastErr == nil {
+			if attempt > 1 {
+				f.logger.Info(fmt.Sprintf("Succeeded on attempt %d after previous failures", attempt+1))
+			}
+			break
```
g-gaston, line 84: should this be return? at this point we have succeeded right?

test/e2e/suite/flake_attempt.go
g-gaston, line 16: nit wdty about
```suggestion
type FlakyCode struct {
```

so it reads something like `flakyCode.It(ctx, "does something", maxRetries`

test/e2e/suite/nodeadm_test.go
```diff
@@ -243,40 +243,66 @@ var _ = Describe("Hybrid Nodes", func() {
...
+									Expect(ec2.WaitForEC2InstanceRunning(ctx, test.ec2Client, instance.ID)).To(Succeed(), "EC2 Instance should have been reached Running status")
+									_, err := verifyNode.WaitForNodeReady(ctx)
+									if err != nil {
+										// an ec2 node is considered impaired if the reachability health check fails
+										// in these cases we want to retry by deleting the instance and recreating it
+										// to avoid failing tests due to instance boot issues that are not related
+										// to nodeadm or the test itself
+										isImpaired, oErr := ec2.IsEC2InstanceImpaired(ctx, test.ec2Client, instance.ID)
+										if oErr != nil {
+											Expect(err).NotTo(HaveOccurred(), "should describe instance status")
```
g-gaston, line 292: should this be
```suggestion
											Expect(oErr).NotTo(HaveOccurred(), "should describe instance status")
```

test/e2e/suite/flaky_code.go
```diff
@@ -0,0 +1,150 @@
...
+	panicableError error
+	retryableError error
+	shouldRetry    bool
+	testGoMega     gomega.Gomega
+}
+
+func newRetryable(attempt int, maxAttempts int) *retryable {
+	retry := &retryable{}
+	retry.testGoMega = gomega.NewGomega(func(message string, callerSkip ...int) {
+		retry.shouldRetry = attempt < maxAttempts-1
```
g-gaston, line 39: not a big deal but i think this makes it more clear that `shouldRetry` doesn't ever change in this struct
```suggestion
	retry := &retryable{
	    shouldRetry = attempt < maxAttempts-1
	}
	retry.testGoMega = gomega.NewGomega(func(message string, callerSkip ...int) {
```

test/e2e/suite/flaky_code.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// if there is no panic, we don't need to store the error
+	// and the test can continue
+	if e == nil {
+		return
+	}
+
+	err, ok := e.(types.GinkgoError)
+	if !ok {
+		err = types.GinkgoError{
```
g-gaston, line 73: should this maybe be always panicable no matter what `shouldRetry` says?

test/e2e/ec2/instance.go
```diff
@@ -163,6 +163,27 @@ func WaitForEC2InstanceRunning(ctx context.Context, ec2Client *ec2.Client, insta
...
+func IsEC2InstanceImpaired(ctx context.Context, ec2Client *ec2.Client, instanceID string) (bool, error) {
+	describeStatusOutput, err := ec2Client.DescribeInstanceStatus(ctx, &ec2.DescribeInstanceStatusInput{
+		InstanceIds:         []string{instanceID},
+		IncludeAllInstances: aws.Bool(true),
+	})
+	if err != nil {
+		return false, fmt.Errorf("describing instance status %s: %w", instanceID, err)
+	}
+	for _, status := range describeStatusOutput.InstanceStatuses[0].SystemStatus.Details {
+		if status.Name == "reachability" {
```
abhay-krishna, line 175: I'm curious why we are only checking for the name/type of status and not the status itself, like `passed` or `failed`? Also the `InstanceStatus` and `SystemStatus` response objects also have a `Status` field outside of the `Details` field, which provides the overall status of the instance or system health taking into account all the indivudual health reports in the `Details` field. This Status field takes the following values:`ok | impaired | insufficient-data | not-applicable | initializing`. Maybe we could just check for `ok` here instead of drilling into each of the reports details.

test/e2e/ec2/instance.go
```diff
@@ -163,6 +163,27 @@ func WaitForEC2InstanceRunning(ctx context.Context, ec2Client *ec2.Client, insta
...
+			return true, nil
+		}
+	}
+	for _, status := range describeStatusOutput.InstanceStatuses[0].InstanceStatus.Details {
+		if status.Name == "reachability" {
```
abhay-krishna, line 180: Same comment as above

test/e2e/ec2/instance.go
jaxesn: Totally, we could. Each of the checks in instancesStatus and systemStatus roll up to that overall status you are referring to.

My thinking was to limit it down to what weve seen happen and we know is not due to our code. Then if new cases happen we can investigate them on a case by case, making sure we can confirm its really an ec2 thing that we should be retrying for.

abhay-krishna: That makes sense but even in the limited check case, we need to check for `status.Name == "reachability"` and `status.Status == "failed"` right? The name check alone could lead to false positives, since the status could be `passed`.

g-gaston (approved):

PR #387, opened 2025-03-08
improves e2e pod identity debugging

jaxesn:
*Issue #, if available:*

*Description of changes:*

Ran into a few issues while testing locally with the new pod identity test:

- We want e2e-test cleanup to be able to be run multiple times and not fail out due to resources already being cleaned up.  I removed the code in the stack.go:delete which relied on the stack information to find the bucket. Instead move the code from nodeadm_test into a helper, similar to the way we handle the jumpbox, and use that during the test and the delete.
- The test was failing for me locally, which ended up being due to having left around buckets from previous test clusters, i use the same cluster name.  I added more steps to the test to try and pinpoint where the issue was. This ended up being that since i had multiple buckets, the code would pick one and that one it picked, the test file had been deleted.  I added an error in the case that it finds multiple buckets.
- Changed the runPod method to return the stdout/stderr even if there is an error so we can output it in the returned error. This was helpful to see the actual s3 error that was returned due to the missing file.
-  changed the waitForPodRunnign method to accept a ListOptions instead of a name so we could use it to wait for the demonset pod on the actual node we were testing on.
- add context to each of the returned errors

*Testing (if applicable):*

### DIFF
--- a/test/e2e/addon/podidentityverify.go
+++ b/test/e2e/addon/podidentityverify.go
@@ -62,17 +62,21 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
 	defer cancel()
 
 	if err := podIdentityAddon.WaitUtilActive(timeoutCtx, v.EKSClient, v.Logger); err != nil {
-		return err
+		return fmt.Errorf("waiting for pod identity add-on to be active: %w", err)
 	}
 
 	v.Logger.Info("Check if daemon set exists", "daemonSet", podIdentityDaemonSet)
 	if _, err := kubernetes.GetDaemonSet(ctx, v.Logger, v.K8S, "kube-system", podIdentityDaemonSet); err != nil {
-		return err
+		return fmt.Errorf("getting daemon set %s: %w", podIdentityDaemonSet, err)
 	}
 
 	node, err := kubernetes.WaitForNode(ctx, v.K8S, v.NodeIP, v.Logger)
 	if err != nil {
-		return err
+		return fmt.Errorf("waiting for node %s to be ready: %w", v.NodeIP, err)
+	}
+
+	if err := kubernetes.WaitForDaemonSetPodToBeRunning(ctx, v.K8S, "kube-system", podIdentityDaemonSet, node.Name, v.Logger); err != nil {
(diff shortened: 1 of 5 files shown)


junpengdev (approved):

test/e2e/addon/podidentityverify.go
```diff
@@ -62,17 +62,21 @@ func (v VerifyPodIdentityAddon) Run(ctx context.Context) error {
 	defer cancel()
 
 	if err := podIdentityAddon.WaitUtilActive(timeoutCtx, v.EKSClient, v.Logger); err != nil {
-		return err
+		return fmt.Errorf("failed to wait for pod identity add-on to be active: %w", err)
```
g-gaston, line 65:
```suggestion
		return fmt.Errorf("waiting for pod identity add-on to be active: %w", err)
```

test/e2e/cluster/stack.go
```diff
@@ -310,6 +301,23 @@ func (s *stack) delete(ctx context.Context, clusterName string) error {
 	return nil
 }
 
+func (s *stack) deletePodIdentityS3Bucket(ctx context.Context, clusterName string) error {
```
g-gaston, line 304:
```suggestion
func (s *stack) emptyPodIdentityS3Bucket(ctx context.Context, clusterName string) error {
```

test/e2e/peered/podidentitybucket.go
g-gaston, line 1: nit: i would probably put this in the addon package since there is nothing about this that is exclusive of hybrid running on top of ec2

test/e2e/peered/podidentitybucket.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		}
+	}
+
+	if len(foundBuckets) > 1 {
+		return "", fmt.Errorf("found multiple pod identity buckets for cluster %s: %v", cluster, foundBuckets)
+	}
+
+	if len(foundBuckets) == 0 {
+		return "", errors.New(PodIdentityBucketNotFound)
```
g-gaston, line 58: Not: I think a more common pattern in go is to make the error itself a package level var. then from the consumer you can use `errors.Is` to check for it

PR #402, opened 2025-03-17
adds retries around install artifacts if checksums fail

jaxesn:
*Issue #, if available:*

*Description of changes:*

When adding gpg validation to ssm, I added retries around both the downloading and the checksum validation to catch cases where the checksum fails due to a bad download.  I actually saw something like this in 1 test run in my devstack.

This PR changes the remaining artifact downloading to match this pattern where checksum validation can trigger a retry.

Note:

The new unit tests are basically the same between the different artifacts.

*Testing (if applicable):*

### DIFF
--- a/internal/artifact/source.go
+++ b/internal/artifact/source.go
@@ -1,6 +1,7 @@
 package artifact
 
 import (
+	"fmt"
 	"hash"
 	"io"
 )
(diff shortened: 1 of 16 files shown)


g-gaston (commented):
i like it
Let's merge as is without refactor, but i think now that we understand how the installation code of the diff artifacts evolve together, we should look into making some of this code re-usable. The fact that you wrote a mini test framework that we reuse in multiple unit tests makes me think some of this code should be in a central place and called by the diff installations.

g-gaston (approved):

PR #418, opened 2025-03-24
add e2e-label to kube node to detect joining

jaxesn:
*Issue #, if available:*

*Description of changes:*

Adds a new label to the created kubenetes nodes in the e2e tests.  This label will be used to detect when the node joins the instead of the internal IP.  

We have seen cases where it appears that the node joins and gets ready, but it is never found by the internalip.  This does include a check when the node joins, validating the joined internal IP with the instance IP.  For now, this will still fail the test, but may gave us a clue into what is causing this discrepancy. 

The instanceName is used, instead of the nodeName because the created ec2.Instance via peeredNode, has the instanceName instead of the nodeName. Using the instanceName made it easier to grab the correct value in the various places throughout the test.  Not opposed to use the nodeName instead.

*Testing (if applicable):*

### DIFF
--- a/cmd/e2e-test/node/create.go
+++ b/cmd/e2e-test/node/create.go
@@ -135,7 +135,7 @@ func (c *create) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		}
 	}
 
-	instance, err := node.Create(ctx, &peered.NodeSpec{
+	peerdNode, err := node.Create(ctx, &peered.NodeSpec{
 		InstanceName:   c.instanceName,
 		NodeK8sVersion: cluster.KubernetesVersion,
 		NodeName:       c.instanceName,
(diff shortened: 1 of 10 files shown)


test/e2e/kubernetes/kubernetes.go
```diff
@@ -40,27 +40,28 @@ const (
...
 			if consecutiveErrors > 3 {
 				return false, err
 			}
-			logger.Info("Retryable error listing nodes when looking for node with IP. Continuing to poll", "internalIP", internalIP, "error", err)
+			logger.Info("Retryable error listing nodes when looking for node with name. Continuing to poll", "nodeName", nodeName, "error", err)
 			return false, nil // continue polling
 		}
 		consecutiveErrors = 0
-		if node != nil {
+		if node != nil && GetNodeInternalIP(node) != "" {
```
g-gaston, line 59: should this be in `WaitForNodeReady`?
I feel like this method should just wait until a matching node exists, not until a matching node has an IP

g-gaston (approved):

PR #419, opened 2025-03-24
Add proxy flags to IAM-RA

vignesh-goutham:
*Description of changes:*
IAM-RA uses aws_signing_helper binary which requires an explicit `--with-proxy` flag to be set if we expect the binary to consume the proxy env vars. This changes checks if proxy env vars are set and adds the flag.


pokearu (approved):
/lgtm
/approve

PR #421, opened 2025-03-25
adds new integration test mocking eks api

jaxesn:
*Issue #, if available:*

*Description of changes:*

When working on a new debug command validation, I wanted to validate that config enrichment worked during that command flow.  I realized we did not have any integration test that mocked the eks api.  

This adds a new test, more so as an example since our e2e tests validation config enrichment, that shows how to mock the eks api, with iam-ra, and validates that if the ca/remoteconfig/apiendpoint arent supplied via the config, they are pulled from the eks api.

We needed a [fix](https://github.com/getmoto/moto/pull/8710) to moto due to moto's handling of timestamp types for eks clusters, but there was already a PR open which merged today.  For now, the second commit, installs moto from a specific commit has. Once there is a new release we can revert that commit.

I originally implemented this by patching moto after install doing something like this.  This could be helpful in the future if we run into other cases with moto where we need to temporarily patch while waiting on a release.  We should have a way to handle this and not be blocked on their release timeline.

```
RUN pip install --user 'moto[server] @ git+https://github.com/getmoto/moto.git@0637c5c46a1f49a1ce3defff8e508d6f140344f8'

COPY test/integration/infra/moto-patches/*.patch /moto-patches/
RUN patch -p1 -d /root/.local/lib/python*/site-packages/ < /moto-patches/8710.patch
```

*Testing (if applicable):*

### DIFF
--- a/test/integration/docs/patching-moto.md
+++ b/test/integration/docs/patching-moto.md
@@ -4,9 +4,19 @@ This document describes how to temporarily patch moto in the integration test Do
 
 ## Overview
 
-Sometimes we need to carry patches for moto while waiting for upstream releases. This document outlines the process for creating and applying patches to moto in our integration test environment.
+Sometimes we need to switch to an unreleased moto or carry patches for moto while waiting for upstream releases. This document outlines the process for creating and applying patches to moto in our integration test environment.
 
-## Steps to Patch Moto
+## Steps to install unreleased moto
+
+1. Update the Dockerfile to pip install from a specific commit:
+   ```dockerfile
+   # Install moto from specific commit
+   RUN pip install --user 'moto[server] @ git+https://github.com/getmoto/moto.git@<commit-hash>'
+   ```
+
+2. Document the reason for using the specific commit in a comment above the installation line.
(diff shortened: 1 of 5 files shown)


test/integration/docs/patching-moto.md
```diff
@@ -0,0 +1,41 @@
...
+1. Create a `moto-patches` directory in `test/integration/infra/` if it doesn't exist:
+   ```bash
+   mkdir -p test/integration/infra/moto-patches
+   ```
+
+2. Generate a patch file, ideally open a PR upstream with the changes and store in the `moto-patches` folder.
+
+3. Update the Dockerfile to apply the patch:
+   ```dockerfile
+   # Install moto from specific commit
```
achen1210, line 20: this can be used separately from the patching right if its already in moto's main branch? it might be more clear if this is moved to before step 1 i.e. if change is in moto upstream main but not yet released, grab moto from commit, else start from moto at a specific commit and add more patches as needed.

achen1210 (approved):

PR #426, opened 2025-03-27
removes aws config and ssm symlink on uninstall

jaxesn:
*Issue #, if available:*

*Description of changes:*

Weve seen an issue in our e2e tests where when a node rejoins, after running uninstall, ssm can sometimes take a few iterations of its cred refresh loop to get the new creds for the newly activated hybrid node.  While this is happening, kubelet has already exec'd iam-auth to get creds which are for the previous hybrid activation, which is still valid.  These valid creds are cached, but the kubelet cant use them to register itself because its using the old nodes creds.

This removes the aws config file and symlink that we create for ssm to ensure that on uninstall, the creds are all removed and cant be reused after reinstalling. 

*Testing (if applicable):*

### DIFF
--- a/internal/flows/uninstall.go
+++ b/internal/flows/uninstall.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/aws/aws-sdk-go-v2/config"
+	awsSsm "github.com/aws/aws-sdk-go-v2/service/ssm"
 	"go.uber.org/zap"
 
 	"github.com/aws/eks-hybrid/internal/containerd"
@@ -23,7 +25,6 @@ import (
 const eksConfigDir = "/etc/eks"
 
 type (
-	SSMUninstall func(ctx context.Context, logger *zap.Logger, pm ssm.PkgSource) error
 	CNIUninstall func() error
 )
 
@@ -31,7 +32,6 @@ type Uninstaller struct {
 	Artifacts      *tracker.InstalledArtifacts
 	DaemonManager  daemon.DaemonManager
 	PackageManager *packagemanager.DistroPackageManager
-	SSMUninstall   SSMUninstall
 	Logger         *zap.Logger
 	CNIUninstall   CNIUninstall
 }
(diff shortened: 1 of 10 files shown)


internal/ssm/install.go
```diff
@@ -202,7 +202,15 @@ func DeregisterAndUninstall(ctx context.Context, logger *zap.Logger, pkgSource P
...
 	}
 
-	return os.RemoveAll(symlinkedAWSConfigPath)
+	if err := os.RemoveAll(symlinkedAWSConfigPath); err != nil {
+		return errors.Wrapf(err, "failed to uninstall ssm aws config symlink")
+	}
+
+	if err := os.RemoveAll(defaultAWSConfigPath); err != nil {
+		return errors.Wrapf(err, "failed to uninstall ssm aws config")
+	}
```
g-gaston, line 211: tests? given the code is trivial, some assertions in the existing integration tests are probably enough I think
Unit test this is a bit difficult if we want to keep the paths in constants

internal/ssm/registration.go
```diff
@@ -0,0 +1,143 @@
...
+	return true, nil
+}
+
+// RegistrationFilePath returns the path to the SSM registration file
+// If installRoot is not set, it will return the path starting from the disk root
+func (r *SSMRegistration) RegistrationFilePath() string {
+	return filepath.Join(r.installRoot, registrationFilePath)
+}
+
+func (r *SSMRegistration) SSMClient(ctx context.Context) (SSMClient, error) {
```
g-gaston, line 127: I would just offer a method that returns the region
That can use to build the client or for other things but it decouples reading the region from a registration from building a client for a particular region

internal/ssm/registration.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/aws/aws-sdk-go-v2/config"
+	awsSsm "github.com/aws/aws-sdk-go-v2/service/ssm"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+const registrationFilePath = "/var/lib/amazon/ssm/registration"
+
+type SSMRegistration struct {
+	installRoot string
```
g-gaston, line 19: If we are gonna introduce an interface for registration and mock it in the tests, do we even need this? I see that you use this from the mock itself, but that's kind of weird. We are testing the implementation of its methods by reusing it to implement a mock. vs letting the code using it directly.

I still like this field and testing the real deletion though. Maybe what we can do is avoid mocking the whole registration? It seems like the only reason we need to mock it is because it returns a client. So what if we just inject the client as another dependency to de-register? we can then only mock the client and use `SSMRegistration` directly

internal/flows/uninstall.go
```diff
@@ -70,7 +70,25 @@ func (u *Uninstaller) uninstallDaemons(ctx context.Context) error {
...
+		opts := []func(*config.LoadOptions) error{}
+		if region != "" {
+			opts = append(opts, config.WithRegion(region))
+		}
+
+		awsConfig, err := config.LoadDefaultConfig(ctx, opts...)
+		if err != nil {
+			return err
+		}
+		fmt.Println("awsConfig", awsConfig)
```
g-gaston, line 84:
```suggestion
```

g-gaston (dismissed):

g-gaston (approved):

PR #428, opened 2025-03-27
add retries for node validations

tatlat:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/internal/node/pods.go
+++ b/internal/node/pods.go
@@ -11,10 +11,10 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/yaml"
-
-	"github.com/aws/eks-hybrid/internal/kubelet"
 )
 
 type podFilter func([]corev1.Pod) ([]corev1.Pod, error)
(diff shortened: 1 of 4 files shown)


internal/node/pods.go
```diff
@@ -92,13 +94,25 @@ func getPodsOnNode(ctx context.Context, nodeName string) ([]corev1.Pod, error) {
...
 	}
 
-	pods, err := clientset.CoreV1().Pods("").List(ctx,
-		metav1.ListOptions{
-			FieldSelector: fmt.Sprintf("spec.nodeName=%s", nodeName),
-		},
-	)
+	var pods *corev1.PodList
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, 30*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
```
jaxesn, line 99: lets move to the timeouts to constants

internal/node/pods.go
```diff
@@ -92,13 +93,25 @@ func getPodsOnNode(ctx context.Context, nodeName string) ([]corev1.Pod, error) {
...
 	}
 
-	pods, err := clientset.CoreV1().Pods("").List(ctx,
-		metav1.ListOptions{
-			FieldSelector: fmt.Sprintf("spec.nodeName=%s", nodeName),
-		},
-	)
+	var pods *corev1.PodList
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, nodeValidationInterval, nodeValidationTimeout, true, func(ctx context.Context) (bool, error) {
```
g-gaston, line 98: good idea
could we test this?

internal/node/validations.go
```diff
@@ -3,17 +3,24 @@ package node
...
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/aws/eks-hybrid/internal/kubelet"
 )
 
 const defaultStaticPodManifestPath = "/etc/kubernetes/manifest"
 
+const (
+	nodeValidationInterval = 30 * time.Second
+	nodeValidationTimeout  = 5 * time.Minute
```
g-gaston, line 21: Are this random values or we have some reason behind them?

tatlat: they're random

tatlat: Those were random placeholders while I test. I'm going to change it to 5 second intervals and 1 minute timeout. Since it stops after 5 consecutive errors, 1 minute makes sense since it allows for up to 6 retries assuming there's no delay in getting nodes. 5 seconds should be enough time for transient issues to resolve without polling too frequently.

tatlat: If it turns out that's not enough, we can increase the duration in the future.

internal/node/validations.go
```diff
@@ -74,5 +87,24 @@ func getCurrentNode(ctx context.Context) (*v1.Node, error) {
...
+
+func getNode(ctx context.Context, nodeName string, clientset kubernetes.Interface) (*v1.Node, error) {
+	var node *v1.Node
+	var err error
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, nodeValidationInterval, nodeValidationTimeout, true, func(ctx context.Context) (bool, error) {
+		node, err = clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+		if err != nil {
+			consecutiveErrors += 1
+			if consecutiveErrors > 4 {
```
achen1210, line 101: maybe make number of retries a constant and then set to 5 using >=?

achen1210 (approved):

Issue #413, opened 2025-03-19
Improve nodeadm kubelet cert validations


csplinter:
**What would you like to be added**:
1. Make sure `nodeadm uninstall` removes the kubelet certs
2. Add a validation to `nodeadm init` to check for the existence of old / invalid kubelet certs
3. Add a kubelet cert validation check to `nodeadm debug`

**Why is this needed**:
During early phases of using hybrid nodes, users are likely to trial / error test when things don't work properly on the first try. This can result in old kubelet certs being present on a hybrid node, which can lead to issues that are hard to debug and fix.

For example, if you have old kubelet certs for an old cluster on your node, you may be able to join your nodes to a new EKS cluster, but you won't be able to pull the logs because the API server can't connect to the kubelet API.

Example error if this is the case:
```
Error from server: Get [https://<ip>:10250/containerLogs/kube-system/coredns-<id>/coredns](https://urldefense.com/v3/__https:/<ip>:10250/containerLogs/kube-system/coredns-<id>/coredns__;...): tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kubernetes")
```

DLeich:
+1 We ran into the exact same situation while trialing hybrid nodes.  This would be an important enhancement to ensure a clean setup when reinstalling.

PR #430, opened 2025-03-28
add kubelet server cert validation to the debug command

jaxesn:
*Issue #, if available:*

*Description of changes:*

This implements part of #413 adding the cert validations to the debug command.

*Testing (if applicable):*


internal/kubernetes/cert.go
```diff
@@ -0,0 +1,51 @@
...
+func (v KubeletCertificateValidator) Run(ctx context.Context, informer validation.Informer, node *api.NodeConfig) error {
+	name := "kubernetes-kubelet-certificate"
+	var remedationErr error
+	informer.Starting(ctx, name, "Validating kubelet server certificate")
+	defer func() {
+		informer.Done(ctx, name, remedationErr)
+	}()
+
+	if err := hybrid.ValidateKubeletCert(v.logger, v.installRoot, node.Spec.Cluster.CertificateAuthority); err != nil {
+		remedationErr = validation.WithRemediation(err, err.Remediation)
```
g-gaston, line 46: instead of doing this I would just make `ValidationError` implement the `Remediable` interface

internal/node/hybrid/kubelet_cert_validator.go
```diff
@@ -13,51 +13,108 @@ import (
...
+	return e.Message
+}
+
+func (e *ValidationError) Unwrap() error {
+	return e.Cause
+}
+
 // ValidateKubeletCert checks if there is an existing kubelet certificate and validates it against the cluster's CA
-func ValidateKubeletCert(logger *zap.Logger, installRoot string, ca []byte) error {
+func ValidateKubeletCert(logger *zap.Logger, installRoot string, ca []byte) *ValidationError {
```
g-gaston, line 37: not a blocker
In go is a bit more common to just return the `error` interface and offer exported methods to check the "type" of error

That allows to keep a common/consistent method signature across methods, to not add complexity for most callers while still giving the option to check the type of errors to the callers who care

you could just have

```go

func hybrid.IsDateValidationError(err error) bool

func hybrid.IsNoCertError(err error) bool
```

internal/kubernetes/cluster.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func NewClusterProvider(config aws.Config) ClusterProvider {
+	return &clusterProvider{
+		aws: config,
+	}
+}
+
+// ReadClusterDetails returns ClusterDetails with caching, delegating to eks.ReadClusterDetails for the actual API call
+func (p *clusterProvider) ReadClusterDetails(ctx context.Context, node *api.NodeConfig, informer validation.Informer) (*api.ClusterDetails, error) {
```
g-gaston, line 29: personal preference, but i would suggest making this "validation agnostic" but keeping the informer out of here
That might lead to a bit of duplication among the different validations (calling the informer on error and what not) but I find that cleaner since it keeps all the validation framework handling in the validation itself and this just becomes a read+cache. It also allows each validation to handle an error of this method however the want.
The remediation i think it's fine to have it here since it transparent to the method signature

internal/kubernetes/cert.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/aws/eks-hybrid/internal/api"
+	"github.com/aws/eks-hybrid/internal/node/hybrid"
+	"github.com/aws/eks-hybrid/internal/validation"
+)
+
+type KubeletCertificateValidator struct {
+	// InstallRoot is optionally the root directory of the installation
+	// If not provided, the installRoot will be empty representing the root (/) of the fs
+	installRoot     string
```
g-gaston, line 14: what if instead of install root we pass the path to the certificate?
I find that a bit more clear and it avoids path manipulation down the stack

internal/kubernetes/cert.go
```diff
@@ -0,0 +1,54 @@
...
+	}()
+
+	cluster, err := v.clusterProvider.ReadClusterDetails(ctx, node, informer)
+	if err != nil {
+		remedationErr = err
+		return remedationErr
+	}
+
+	if err := hybrid.ValidateKubeletCert(v.installRoot, cluster.CertificateAuthority); err != nil {
+		remedationErr = hybrid.AsValidationError(err)
```
g-gaston, line 49: why do we need this? In this case I would expect `hybrid.ValidateKubeletCert` to return a remediable error when there is a known remediation so we can just return whatever it gives us

g-gaston (approved):

PR #432, opened 2025-03-28
Set timeoutSeconds and periodSeconds for readiness probe

junpengdev:
*Issue #, if available:*

*Description of changes:*

Fine tune timeoutSeconds and periodSeconds to improve reliability for pod identity test.
Default timeoutSeconds is 1 so our probe considers the command execution `aws sts get-caller-identity` failed if it does not receive response within 1 seconds. However this is a remote call, which can takes more than 1 seconds to complete. Without this change, current setting for probe causes pod in `NotReady` state.   

*Testing (if applicable):*
Manually triggered the CI in dev stack and it succeeded multiple times. 


jaxesn (approved):
I ran a version of this in my devstack a couple times as well and saw success as well.

PR #446, opened 2025-04-03
adds context deadlines for run-e2e flow

jaxesn:
*Issue #, if available:*

*Description of changes:*

We want to try and ensure cleanup and reporting always happens.  There are cases where the test setup/tests themselves can run too long and trigger the overall canary timeout. In those cases we do not get the result json nor do we cleanup the infra.  

This adds deadlines to the various contexts to try and ensure we always have time to run the final steps.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/run/e2e.go
+++ b/test/e2e/run/e2e.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/go-logr/logr"
@@ -30,6 +31,7 @@ const (
 	phaseNameSetupTestInfrastructure = "SetupTestInfrastructure"
 	phaseNameUploadArtifactsToS3     = "UploadArtifactsToS3"
 	phaseNameWriteTestConfigs        = "WriteTestConfigs"
+	reportingBuffer                  = time.Minute * 2
 	testCleanupLogFile               = "cleanup-output.log"
 	testGinkgoOutputLog              = "ginkgo-output.log"
 	testSetupLogFile                 = "setup-output.log"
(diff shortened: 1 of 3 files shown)


test/e2e/run/e2e.go
```diff
@@ -174,14 +180,17 @@ func (e *E2E) preTestSetup() ([]Phase, error) {
 }
 
 func (e *E2E) runTests(ctx context.Context) ([]Phase, error) {
+	// save 2 minutes for reporting/s3 uploading
+	deadline, _ := ctx.Deadline()
+	ctx, testCancelFunc := context.WithDeadline(ctx, deadline.Add(-reportingBuffer))
```
g-gaston, line 185: What if we do this in the top level method so we dont need to extract the deadline for the context? It also keeps the timeout handling at the top level

test/e2e/run/runner.go
```diff
@@ -54,13 +60,18 @@ func (e *E2ERunner) Run(ctx context.Context) ([]Phase, error) {
 		}
 	}()
 
-	setupErr := e.setupTestInfrastructure(ctx)
+	// save 5 minutes for cleanup
+	deadline, _ := ctx.Deadline()
```
g-gaston, line 64: What happens here if the context doesn't have a deadline?

test/e2e/run/runner.go
```diff
@@ -89,6 +100,12 @@ func (e *E2ERunner) executeTests(ctx context.Context) error {
 		return fmt.Errorf("getting working directory: %w", err)
 	}
 
+	deadline, _ := ctx.Deadline()
+	timeout := (time.Until(deadline) - ginkgoCleanupBuffer).Round(time.Second)
```
g-gaston, line 104: Same here
Could we do this at the top level?

test/e2e/run/runner.go
```diff
@@ -89,6 +100,12 @@ func (e *E2ERunner) executeTests(ctx context.Context) error {
 		return fmt.Errorf("getting working directory: %w", err)
 	}
 
+	deadline, _ := ctx.Deadline()
```
g-gaston, line 103: Handle no deadline?

g-gaston (approved):

PR #450, opened 2025-04-04
ignore NoSuchTagSet when finding s3 buckets to cleanup

jaxesn:
*Issue #, if available:*

*Description of changes:*

We are seeing cases where the pre test sweeper is running to try and find old resources. Since s3 buckets require a second call to get the tags, we currently check for bucketnotexist when getting tags to handle the case where the bucket is deleted between the call of getting the buckets and the call to get the tags.  This changes now handles the opposite case where the bucket is created but the tagging is done yet.

Also added a check for a 403 when trying to delete the cluster, which happens when using the CI role which is limited by tags if the cluster is already deleted.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/cleanup/eks.go
+++ b/test/e2e/cleanup/eks.go
@@ -61,10 +61,21 @@ func (c *EKSClusterCleanup) DeleteCluster(ctx context.Context, clusterName strin
 	_, err := c.eksClient.DeleteCluster(ctx, &eks.DeleteClusterInput{
 		Name: aws.String(clusterName),
 	})
+
+	if err != nil && errors.IsAwsError(err, "AccessDeniedException") {
+		// if the cluster deleted, the role policy may return a 403 since its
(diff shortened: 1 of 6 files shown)


test/e2e/cleanup/eks.go
```diff
@@ -61,10 +61,21 @@ func (c *EKSClusterCleanup) DeleteCluster(ctx context.Context, clusterName strin
 	_, err := c.eksClient.DeleteCluster(ctx, &eks.DeleteClusterInput{
 		Name: aws.String(clusterName),
 	})
+
+	if err != nil && errors.IsAwsError(err, "AccessDeniedException") {
+		// if the cluster deleted, the role policy may return a 403 since its
```
drewvanstone, line 66:
```suggestion
		// if the cluster has been deleted, the role policy may return a 403 since its
```

PR #454, opened 2025-04-04
improves run-e2e flow and logging

jaxesn:
*Issue #, if available:*

*Description of changes:*

Add more logging local to where the error may have happened to make tracking logs easier.  Improves use of the Phase concept and phaseComplete method by returning Phase, which already included the error, instead of both Phase, err.  The error return is now limited to the main entry point of the run-e2e command.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/run/e2e.go
+++ b/test/e2e/run/e2e.go
@@ -108,29 +108,24 @@ func (e *E2E) Run(ctx context.Context) (E2EResult, error) {
 	}
 
 	// After this point we are going to run as much as we can regardeless of potential errors
-	// and will collect all errors to return them at the end
-	allErrors := []error{}
-
-	runTestsPhases, err := e.runTests(ctx)
-	if err != nil {
-		allErrors = append(allErrors, err)
-	}
+	runTestsPhases := e.runTests(ctx)
 	phases = append(phases, runTestsPhases...)
 
-	e2eResult, err := e.parseReport(ctx)
-	phases, err = phaseCompleted(phases, phaseNameParseReport, "parsing report", err)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


test/e2e/run/e2e.go
```diff
@@ -113,34 +113,35 @@ func (e *E2E) Run(ctx context.Context) (E2EResult, error) {
...
+	phases = append(phases, parsePhases...)
+
+	uploadPhases := e.uploadArtifactsToS3(ctx)
+	phases = append(phases, uploadPhases...)
 
 	e2eResult.Phases = phases
-	return e2eResult, errors.Join(allErrors...)
+	var allErrors error
+	for _, phase := range phases {
+		allErrors = errors.Join(allErrors, phase.Error)
```
g-gaston, line 142: should we check if err != nil? i'm assuming that errors Join only adds non nil errors but it seems more explicit to maybe do it here?

test/e2e/run/e2e.go
```diff
@@ -192,31 +193,54 @@ func (e *E2E) runTests(ctx context.Context) ([]Phase, error) {
...
-		SkipCleanup:     e.SkipCleanup,
 		SkippedTests:    e.SkippedTests,
 	}
 	return runner.Run(ctx)
 }
 
+func (e *E2E) runCleanup(ctx context.Context) []Phase {
+	if e.SkipCleanup {
+		e.Logger.Info("Skipping cluster and infrastructure cleanup via stack deletion")
+		return []Phase{}
```
g-gaston, line 204:
```suggestion
		return nil
```

g-gaston (approved):

PR #460, opened 2025-04-08
adds error message logging before ginkgo running defer cleanups

jaxesn:
*Issue #, if available:*

*Description of changes:*

Currently when an Expect failures in a ginkgo test the error is not actually logged until the defer cleanups are run. Since we use these extensively it makes our logs a bit hard to follow.

This adds a customer fail handler to log the error message before forwarding to ginkgo's default Fail handler.  The flakyCode block complicated this a bit since it was relying on the global Fail, now the failhandler is passed in from the test so it can call our custom one instead.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/suite/nodeadm_test.go
+++ b/test/e2e/suite/nodeadm_test.go
@@ -22,6 +22,7 @@ import (
 	ssmv2 "github.com/aws/aws-sdk-go-v2/service/ssm"
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
 	clientgo "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -66,7 +67,6 @@ func init() {
 
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
-
 	RunSpecs(t, "E2E Suite")
 }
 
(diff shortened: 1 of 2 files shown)


test/e2e/suite/nodeadm_test.go
```diff
@@ -611,6 +620,27 @@ func (t *peeredVPCTest) newVerifyPodIdentityAddon(nodeName string) *addon.Verify
 	}
 }
 
+func (t *peeredVPCTest) FailHandler(message string, callerSkip ...int) {
```
g-gaston, line 623: nit
```suggestion
// handleFailure is a wrapper around ginkgo.Fail that logs the error message
// immediately after it happens. It doesn't modify gomega's or ginkgo's regular
// behavior.
// We do this to help debug errors when going through the test logs.
func (t *peeredVPCTest) handleFailure(message string, callerSkip ...int) {
```

test/e2e/suite/nodeadm_test.go
```diff
@@ -100,6 +100,8 @@ type peeredVPCTest struct {
 	remoteCommandRunner commands.RemoteCommandRunner
 
 	podIdentityS3Bucket string
+
+	failureMessageLogged bool
```
g-gaston, line 104:
```suggestion
    // failureMessageLogged tracks if a terminal error due to a failed gomega
    // expectation has already been registered and logged . It avoids logging
    // the same multiple times.
	failureMessageLogged bool
```

PR #462, opened 2025-04-09
introduces helpers for common blocks of nodeadm e2e tests

jaxesn:
*Issue #, if available:*

*Description of changes:*

This is an attempt to break out some of the init and upgrade test code that is the same between the two.  Introduced a new TestNode struct to make setup/verify a bit cleaner between the two test.  Moved the peeredVPCTest out into its own file to try and limit nodeadm_test.go to just be the "two" test cases.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/suite/peered_vpc.go
+++ b/test/e2e/suite/peered_vpc.go
@@ -1,0 +1,270 @@
+package suite
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+	ec2v2 "github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/eks"
+	"github.com/aws/aws-sdk-go-v2/service/iam"
+	s3v2 "github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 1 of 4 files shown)


test/e2e/suite/test_node.go
```diff
@@ -0,0 +1,117 @@
...
+type TestNode struct {
+	flakyCode    *FlakyCode
+	instanceName string
+	k8sVersion   string
+	node         *peered.PeerdNode
+	nodeName     string
+	os           e2e.NodeadmOS
+	provider     e2e.NodeadmCredentialsProvider
+	serialOutput peered.ItBlockCloser
+	test         *peeredVPCTest
```
g-gaston, line 27: is there a way to avoid having this in the test node struct?
I don't have anything concrete but it feels weird. Is this because we use a lot of the fields from test

test/e2e/suite/peered_vpc.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// failureMessageLogged tracks if a terminal error due to a failed gomega
+	// expectation has already been registered and logged . It avoids logging
+	// the same multiple times.
+	failureMessageLogged bool
+}
+
+type verificationProvider interface {
+	newPeeredNode() *peered.Node
+	newVerifyNode(nodeName, nodeIP string) *kubernetes.VerifyNode
```
g-gaston, line 84: what if we just move these two methods to testNode? it doesn't seem like they are needed here anymore since the test code now access that functionality through the testNode abstraction.

g-gaston (approved):

PR #463, opened 2025-04-09
adds kubelet cert validation as init preflight

jaxesn:
*Issue #, if available:*

*Description of changes:*

Part 2 of #413 

This adds a validation during init to check for an existing kubelet cert. If the it exists and is valid per the cluster CA, it is allowed to continue. This should cover customers re-runing init on the same node with the same cluster. If the cert is expired or does not validate with the cluster CA an error is returned.

*Testing (if applicable):*


internal/node/hybrid/kubelet_cert_validator.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/aws/eks-hybrid/internal/kubelet"
+)
+
+// ValidateKubeletCert checks if there is an existing kubelet certificate and validates it against the cluster's CA
+func (hnp *HybridNodeProvider) ValidateKubeletCert() error {
```
g-gaston, line 15: not a blocker: I would implement this separate from the hnp, afaik there is no reason why it needs to be tied to it. I know we will want to use this from the `debug` command as well

internal/node/hybrid/kubelet_cert_validator.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	now := time.Now()
+	if now.Before(cert.NotBefore) {
+		return fmt.Errorf("kubelet server certificate is not yet valid (valid from %s). Verify the system time is correct and remove the kubelet server certificate file %s or use --skip kubelet-cert-validation", cert.NotBefore, certPath)
+	}
+
+	if now.After(cert.NotAfter) {
+		return fmt.Errorf("kubelet server certificate has expired (expired at %s). Please remove the kubelet server certificate file %s", cert.NotAfter, certPath)
+	}
```
g-gaston, line 48: i'm not sure if these two belong here
if there is a cert but the cert is expired, the kubelet should request a new one. So I don't think init should fail. I'm not sure about a cert in the future tho.

internal/node/hybrid/kubelet_cert_validator.go
```diff
@@ -0,0 +1,65 @@
...
+		return fmt.Errorf("parsing cluster CA certificate")
+	}
+	// Verify the certificate against the cluster's CA
+	opts := x509.VerifyOptions{
+		Roots:       caPool,
+		CurrentTime: now,
+	}
+
+	if _, err := cert.Verify(opts); err != nil {
+		return fmt.Errorf("kubelet certificate is not valid for the current cluster. Please remove the kubelet server certificate file %s or use --skip kubelet-cert-validation", certPath)
```
g-gaston, line 61:
```suggestion
		return fmt.Errorf("kubelet certificate is not valid for the current cluster. Please remove the kubelet server certificate file %s or use \"--skip kubelet-cert-validation\" if this is expected", certPath)
```

g-gaston (approved):

PR #471, opened 2025-04-11
adds support to override eks endpoint and SPs

jaxesn:
*Issue #, if available:*

*Description of changes:*

When testing agasinst other environments, the endpoint overriding was not complete.  This adds the ability to also overwrite the SPs and passes the endpoint url to the nodeadm call on the node.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/cluster/stack.go
+++ b/test/e2e/cluster/stack.go
@@ -33,6 +33,8 @@ var setupTemplateBody []byte
 
 const (
 	creationTimeParameterKey = "CreationTime"
+	defaultClusterRoleSP     = "eks.amazonaws.com"
+	defaultPodIdentitySP     = "pods.eks.amazonaws.com"
 	stackWaitTimeout         = 7 * time.Minute
 	stackWaitInterval        = 10 * time.Second
 )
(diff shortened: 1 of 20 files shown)


test/e2e/cluster/stack.go
```diff
@@ -108,7 +110,15 @@ func (s *stack) deploy(ctx context.Context, test TestResources) (*resourcesStack
...
-func (s *stack) prepareStackParameters(test TestResources) []types.Parameter {
+func (s *stack) prepareStackParameters(test TestResources, eks EKSConfig) []types.Parameter {
+	clusterRoleSP := defaultClusterRoleSP
+	if eks.ClusterRoleSP != "" {
+		clusterRoleSP = eks.ClusterRoleSP
+	}
+	podIdentitySP := defaultPodIdentitySP
+	if eks.PodIdentitySP != "" {
+		podIdentitySP = eks.PodIdentitySP
+	}
```
g-gaston, line 121: wdyt about setting these defaults where we read the config? that way the struct has the right values all the way in

test/e2e/os/testdata/nodeadm-wrapper.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+AWS_ENDPOINT_URL_EKS={{ .EKSEndpoint }} /tmp/nodeadm-bin "$@"
```
g-gaston, line 2: can we avoid having two scripts by adding some condition in `test/e2e/os/testdata/nodeadm-init.sh` so if the endpoint variable is set then it's passed to nodeadm but if not, then it isn't?

g-gaston (approved):

PR #502, opened 2025-04-18
Add re-tries for getting nodes for deletion

abhinavmpandey08:
*Description of changes:*
Add re-tries for getting nodes for deletion


jaxesn (changes requested):
is this go.sum correct?

jaxesn (approved):
go.sum is just tidy'd

PR #505, opened 2025-04-21
uses shared cred file from iam-ra service when available

jaxesn:
*Issue #, if available:*

*Description of changes:*

Currently when running a single iam-ra test, 16 calls to iam-ra CreateSession are made.  This is because any caching of creds returned from aws_signing_helper has to be done in the calling process.  nodeadm/kubelet/ecr-cred-helper all make their own calls to aws_signing_helper and nodeadm makes multiple depending on the command being run.

This changes `init` to create the iam-ra update systemd unit file before setting up the aws config.  In the case that the user has enabled this service, this will create a `credentials` file with temporary creds.  Within nodeadm we now configure the aws config to look for this service's credentials file.  If the file does not exist, the aws config loader will ignore it and move on to the normal aws config, which has the aws_singing_helper exec configured.

The ecr config created by nodeadm will now also include the credentials file env var and will be used if available, again skipped if the file does not exist.

The kubelet config is left as is, it will exec signing helper itself.  This decision is made because in case the signing_helper service for some reason stops updating the credentials file, the aws config loader will continue to try and use expired creds.  We feel it would be better for kubelet to not rely on this (at least for now).

A single iam-ra e2e test now makes 6 calls to CreateSession
- init -> starting update service which makes 1 call
- init - > starting kubelet service which makes 1 call
- debug -> debug uses the same kubeconfig as kubelet to try and match as close as possible, so this does not rely on the credentials file, resulting in 1 call
- debug -> debug also uses the kubelet kubeconfig to make kube-api requests which end up trigger a new signing helper invoke
- init/kubelet - 2 more calls after node is reset, reboorted and re-init'd

For iam-ra we were using an aws profile name of `hybrid`. This was an issue when trying to use the service generated credential file since it creates them in a `default` profile section. This change includes change the aws profile for iam-ra to be `default`, which matches ssm and the iam-ra service.  

*Note* Just in case someone was using the config file nodeadm creates with aws_signing_helper exec'ing, I now have nodeadm creating 2 profiles, default and hybrid, in the config file.

*Testing (if applicable):*


internal/creds/aws.go
```diff
@@ -32,10 +30,13 @@ func ReadConfig(ctx context.Context, node *api.NodeConfig, opts ...func(*config.
 			awsConfigPath = iamrolesanywhere.DefaultAWSConfigPath
 		}
 
+		// we do not specify the iam-ra credentials file since this is used by the
+		// debug command which we want to match as close as possible to the kubelet which
+		// also does not use the iam-ra credentials file
```
g-gaston, line 35: can we rename the method to something like `ReadConfigAsKubelet`? i want to prevent someone using this one when they should be using the other one

internal/flows/init.go
```diff
@@ -28,6 +28,16 @@ func (i *Initer) Run(ctx context.Context) error {
...
 
+	// in the case of using iam roles anywhere, with EnableCredentialsFile
+	// we want to start the service before configuring aws so we can use the
+	// credentials file to configure aws
+	if !slices.Contains(i.SkipPhases, preprocessPhase) {
+		i.Logger.Info("Configuring Pre-process daemons...")
+		if err := i.NodeProvider.PreProcessDaemon(ctx); err != nil {
+			return err
+		}
+	}
```
g-gaston, line 39: instead of moving this block up, what if we move the "create update service" to the `ConfigureAws` method? i think that's probably where it belong now, in the same way that `ConfigureAws` does the ssm registration to make sure the node has credentials.

internal/iamrolesanywhere/aws_config.tpl
```diff
@@ -1,3 +1,7 @@
 [profile %v]
 region = {{ .Region }}
 credential_process = {{ .SigningHelperBinPath }} credential-process --certificate {{ .CertificatePath }} --private-key {{ .PrivateKeyPath }} --trust-anchor-arn {{ .TrustAnchorARN }} --profile-arn {{ .ProfileARN }} --role-arn {{ .RoleARN }} --role-session-name {{ .NodeName }}{{ if .ProxyEnabled }} --with-proxy{{end}}
+
+[profile hybrid]
```
g-gaston, line 5: nit: if does admit a comment, can we add one here saying this is for backwards compat and nodeadm doesn't use it?

internal/kubelet/image-credential-provider.go
```diff
@@ -83,7 +86,10 @@ func generateImageCredentialProviderConfig(cfg *api.NodeConfig, ecrCredentialPro
 	var imageCredentialProviderTemplate *template.Template
 	if cfg.IsIAMRolesAnywhere() {
 		templateVars.AwsConfigPath = cfg.Spec.Hybrid.IAMRolesAnywhere.AwsConfigPath
+		templateVars.AwsProfile = iamrolesanywhere.ProfileName
+		templateVars.AwsCredentialsPath = iamrolesanywhere.EksHybridAwsCredentialsPath
```
g-gaston, line 90: any way we can inject these in the constructor? just so we don't make the kuebelet package depend on iamrolesanywhere

internal/kubelet/kubeconfig.go
```diff
@@ -117,11 +121,23 @@ func generateKubeconfig(cfg *api.NodeConfig) ([]byte, error) {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to build config from kubeconfig")
+		return nil, errors.Wrap(err, "loading kubeconfig")
 	}
-	return kubernetes.NewForConfig(config)
+	// using the kubelet kubeconfig, but with the iam-ra credentials file set
+	// if the node is running the iam-ra service, this will avoid starting a new session
+	// to make the kuberenetes api calls
+	config.AuthInfos["kubelet"].Exec.Env = []clientcmdapi.ExecEnvVar{
+		{Name: "AWS_SHARED_CREDENTIALS_FILE", Value: iamrolesanywhere.EksHybridAwsCredentialsPath},
```
g-gaston, line 132: I think this method is used by debug, so not sure if this is the right place
what we can do is to accept in this methods closure options

the have another method in the hybrid package like `BuildKubeClient` and that method can call this one with the right option to se the aws creds file

then everywhere that needs a client (except debug) can call this new method. that way it's also obvious we are getting a kube client for the hybrid node case

internal/node/hybrid/aws.go
```diff
@@ -15,9 +14,9 @@ import (
 	"github.com/aws/eks-hybrid/internal/ssm"
 )
 
-const iamRoleAnywhereProfileName = "hybrid"
-
 func (hnp *HybridNodeProvider) ConfigureAws(ctx context.Context) error {
+	configCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
```
g-gaston, line 19: that's going to start counting here instead of just before calling "wait for config"
if the ssm registration plus agent starts takes more than 2 mins, it wont wait for the creds appropiately

internal/node/hybrid/aws.go
```diff
@@ -43,7 +39,8 @@ func (hnp *HybridNodeProvider) ConfigureAws(ctx context.Context) error {
 			return fmt.Errorf("configuring aws credentials with IAM Roles Anywhere: %w", err)
 		}
 
-		awsConfig, err := LoadAWSConfigForRolesAnywhere(ctx, hnp.nodeConfig)
+		hnp.logger.Info("Waiting for AWS config to be available")
+		awsConfig, err := iamrolesanywhere.WaitForAWSConfig(configCtx, hnp.nodeConfig, 2*time.Second, iamrolesanywhere.EksHybridAwsCredentialsPath, hnp.logger)
```
g-gaston, line 43: i know this follows the same pattern we had
but given the wait is conditional to using pod identity or not, i wonder if this wait should be part of the code that configures the `update` service. Like almost as a postflight validation: wait until the service you just started creates the creds file. it also allows us to print the error closer to where it actually happens

internal/iamrolesanywhere/credentials_test.go
```diff
@@ -0,0 +1,110 @@
...
+		expectedError         string
+		expectedAccessKey     string
+		expectedSecretKey     string
+	}{
+		{
+			name:                  "success with credentials file",
+			enableCredentialsFile: true,
+			createCredsFile:       true,
+			credsFileDelay:        2 * time.Millisecond,
+			ctxTimeout:            5 * time.Second,
```
g-gaston, line 33: nit: can we make this faster? like 10 millis or something
these longer tests add up when you want to run all of them
And i dont think waiting for longer provides more coverage

internal/kubelet/kubeconfig.go
```diff
@@ -114,14 +118,50 @@ func generateKubeconfig(cfg *api.NodeConfig) ([]byte, error) {
...
+	// Apply AWS environment variables if provided
+	if len(options.awsEnvVars) > 0 {
+		envVars := make([]clientcmdapi.ExecEnvVar, 0, len(options.awsEnvVars))
+		for name, value := range options.awsEnvVars {
+			envVars = append(envVars, clientcmdapi.ExecEnvVar{
+				Name:  name,
+				Value: value,
+			})
+		}
+		config.AuthInfos["kubelet"].Exec.Env = envVars
```
g-gaston, line 155: nit: worth appending to this in case that there are already some env vars?

g-gaston (approved):

PR #509, opened 2025-04-22
Fix latest paths for prod nodeadm release

vignesh-goutham:
*Description of changes:*
Fix latest path symlinks for nodeadm prod release.

### DIFF
--- a/buildspecs/prod-release-nodeadm.yml
+++ b/buildspecs/prod-release-nodeadm.yml
@@ -25,38 +25,8 @@ phases:
 
   build:
     commands:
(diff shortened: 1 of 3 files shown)


jaxesn (changes requested):
can you add a validation here, maybe in the post_build that just downloads the files to ensure they are what we expect?

jaxesn (commented):
Can you tighten up the cloudfront invalidation to only the latest folder for nodeadm bins and not everything else? 

we may need to do something similar for the manifest release?

jaxesn (approved):

PR #520, opened 2025-04-29
node monitoring agent test

tatlat:
*Issue #, if available:*

*Description of changes:*
This PR adds end-to-end tests for the EKS Node Monitoring Agent addon. Key changes include:

**Added**
- New smoke test suite for addons testing
- Node Monitoring Agent E2E test implementation
- Support for validating node conditions and events created by monitoring agent
- Log collection capabilities to addon struct for debugging failures

*Testing (if applicable):*
Tested with k8s 1.31 and Cilium

### DIFF
--- a/test/e2e/addon/addon.go
+++ b/test/e2e/addon/addon.go
@@ -8,12 +8,14 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/eks"
 	"github.com/aws/aws-sdk-go-v2/service/eks/types"
 	"github.com/go-logr/logr"
+	clientgo "k8s.io/client-go/kubernetes"
 
 	"github.com/aws/eks-hybrid/test/e2e/errors"
 )
 
 type Addon struct {
 	Name          string
+	Namespace     string
 	Cluster       string
 	Configuration string
 }
@@ -54,7 +56,7 @@ func (a Addon) describe(ctx context.Context, client *eks.Client) (*types.Addon, 
 	return describeAddonOutput.Addon, nil
 }
 
-func (a Addon) WaitUtilActive(ctx context.Context, client *eks.Client, logger logr.Logger) error {
+func (a Addon) WaitUntilActive(ctx context.Context, client *eks.Client, logger logr.Logger) error {
 	logger.Info("Describe cluster add-on", "ClusterAddon", a.Name)
 
 	for {
@@ -82,4 +84,33 @@ func (a Addon) WaitUtilActive(ctx context.Context, client *eks.Client, logger lo
 		case <-time.After(backoff):
 		}
 	}
+}
+
+func (a Addon) CreateAndWaitForActive(ctx context.Context, eksClient *eks.Client, k8s clientgo.Interface, logger logr.Logger) error {
+	if err := a.Create(ctx, eksClient, logger); err != nil {
+		return err
+	}
+
+	if err := a.WaitUntilActive(ctx, eksClient, logger); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (a Addon) Delete(ctx context.Context, client *eks.Client, logger logr.Logger) error {
+	logger.Info("Delete cluster add-on", "ClusterAddon", a.Name)
+
+	params := &eks.DeleteAddonInput{
+		ClusterName: &a.Cluster,
+		AddonName:   &a.Name,
+	}
+
+	_, err := client.DeleteAddon(ctx, params)
+
+	if err == nil || errors.IsType(err, &types.ResourceNotFoundException{}) {
+		// Ignore if add-on doesn't exist
+		return nil
+	}
+	return err
 }
--- a/test/e2e/addon/nodemonitoringagent.go
+++ b/test/e2e/addon/nodemonitoringagent.go
@@ -1,0 +1,178 @@
+package addon
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/service/eks"
+	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	clientgo "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	ik8s "github.com/aws/eks-hybrid/internal/kubernetes"
+	"github.com/aws/eks-hybrid/test/e2e/commands"
+	"github.com/aws/eks-hybrid/test/e2e/kubernetes"
+)
+
+const (
+	nodeMonitoringAgentNamespace = "kube-system"
+	nodeMonitoringAgentName      = "eks-node-monitoring-agent"
+	lockupDaemonSetName          = "kernel-soft-lockup"
+	lockupContainerNamespace     = "default"
+
+	nodeMonitoringAgentWaitTimeout = 1 * time.Minute
+)
+
+type NodeMonitoringAgentTest struct {
+	Cluster       string
+	addon         *Addon
+	K8S           clientgo.Interface
+	EKSClient     *eks.Client
+	K8SConfig     *rest.Config
+	Logger        logr.Logger
+	CommandRunner commands.RemoteCommandRunner
+}
+
+func (n *NodeMonitoringAgentTest) Create(ctx context.Context) error {
+	n.addon = &Addon{
+		Cluster:   n.Cluster,
+		Namespace: nodeMonitoringAgentNamespace,
+		Name:      nodeMonitoringAgentName,
+	}
+
+	if err := n.addon.CreateAndWaitForActive(ctx, n.EKSClient, n.K8S, n.Logger); err != nil {
+		return err
+	}
+
+	if _, err := ik8s.GetAndWait(ctx, nodeMonitoringAgentWaitTimeout, n.K8S.AppsV1().DaemonSets(n.addon.Namespace), n.addon.Name, func(ds *appsv1.DaemonSet) bool {
+		return ds.Status.DesiredNumberScheduled == ds.Status.NumberReady
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (n *NodeMonitoringAgentTest) runKernelError(ctx context.Context, nodes *v1.NodeList) error {
+	for _, node := range nodes.Items {
+		ip := kubernetes.GetNodeInternalIP(&node)
+		if ip == "" {
+			return fmt.Errorf("failed to get internal IP for node %s", node.Name)
+		}
+		// This command simulates a kernel error on the node
+		// Node Monitoring Agent should detect this and create an event for SoftLockup
+		// citing KernelReady as the reason.
+		if _, err := n.CommandRunner.Run(ctx, ip, []string{"echo 'watchdog: BUG: soft lockup - CPU#6 stuck for 23s! [VM Thread:4054]' | tee -a /dev/kmsg"}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (n *NodeMonitoringAgentTest) Validate(ctx context.Context) error {
+	nodes, err := ik8s.ListRetry(ctx, n.K8S.CoreV1().Nodes())
+	if err != nil {
+		return err
+	}
+
+	if err := n.runKernelError(ctx, nodes); err != nil {
+		return err
+	}
+
+	for _, node := range nodes.Items {
+		nodeLogger := n.Logger.WithValues("node", node.Name)
+
+		if err := validateNodeConditions(node, nodeLogger); err != nil {
+			return err
+		}
+
+		if err := validateNodeEvents(ctx, n.K8S, node, nodeLogger); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func validateNodeConditions(node v1.Node, nodeLogger logr.Logger) error {
+	foundMonitoringConditions := false
+	for _, condition := range node.Status.Conditions {
+		if isMonitoringAgentCondition(condition) {
+			foundMonitoringConditions = true
+			nodeLogger.Info("Found monitoring agent condition",
+				"type", condition.Type,
+				"status", condition.Status,
+				"message", condition.Message)
+			break
+		}
+	}
+
+	if !foundMonitoringConditions {
+		return fmt.Errorf("no monitoring agent conditions found")
+	}
+
+	return nil
+}
+
+func validateNodeEvents(ctx context.Context, k8s clientgo.Interface, node v1.Node, nodeLogger logr.Logger) error {
+	fieldSelector := fmt.Sprintf("involvedObject.name=%s,involvedObject.kind=Node", node.Name)
+
+	_, err := ik8s.ListAndWait(ctx, nodeMonitoringAgentWaitTimeout, k8s.CoreV1().Events(""), func(events *v1.EventList) bool {
+		foundMonitoringEvents := false
+		for _, event := range events.Items {
+			if isMonitoringAgentEvent(event) {
+				foundMonitoringEvents = true
+				nodeLogger.Info("Found monitoring agent event",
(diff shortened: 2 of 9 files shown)


test/e2e/kubernetes/deployment.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	deploymentSetWaitTimeout   = 3 * time.Minute
```
junpengdev, line 15: `deploymentSet`?

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	return false
+}
+
+func isMonitoringAgentEvent(event v1.Event) bool {
+	return strings.Contains(strings.ToLower(event.Source.Component), nodeMonitoringAgentName)
+}
+
+func (n NodeMonitoringAgentTest) CollectLogs(ctx context.Context) error {
+	return n.addon.FetchLogs(ctx, n.K8S, n.Logger, []string{nodeMonitoringAgentName}, 10)
```
junpengdev, line 254: use `tailLines` const you defined in this package? Also why set the limit to 10 lines? If collecting logs is needed, why not collect all logs?

test/e2e/kubernetes/daemonset.go
```diff
@@ -45,3 +45,30 @@ func GetDaemonSet(ctx context.Context, logger logr.Logger, k8s kubernetes.Interf
 
 	return foundDaemonSet, nil
 }
+
+func WaitForDaemonSetReady(ctx context.Context, logger logr.Logger, k8s kubernetes.Interface, namespace, name string) error {
```
junpengdev, line 49: This function shares a lot of code with `GetDaemonSet`. Is there a way to reduce duplicate?

test/e2e/addon/nodemonitoringagent.go
tatlat: yeah, i guess you're right. I was thinking it might look better in the logs if we kept it short but if there are issues, they're not guaranteed to be in the last 10 lines.

test/e2e/kubernetes/daemonset.go
tatlat: I created a wait for generic resources. Let me know what you think.

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,261 @@
...
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  lockupDaemonSetName,
+							Image: "public.ecr.aws/amazonlinux/amazonlinux:2023-minimal",
```
jaxesn, line 95: this is me being nit picky because it likely doesnt matter since these tests wont be in our canaries, but i would probably still go add this to image to our ecr mirrors and pull it from the ecr account we pull all the other images from.

test/e2e/suite/smoke-tests/addons_test.go
```diff
@@ -0,0 +1,110 @@
+//go:build e2e
+// +build e2e
+
+package addons
```
jaxesn, line 4: i dont fully get the `smoke_tests` folder here? why isnt this just the addons_test?

test/e2e/suite/peered_vpc.go
```diff
@@ -278,6 +278,16 @@ func (t *PeeredVPCTest) NewTestNode(ctx context.Context, instanceName, nodeName,
 	}
 }
 
+func (t *PeeredVPCTest) NewNodeMonitoringAgentTest() *addon.NodeMonitoringAgentTest {
```
g-gaston, line 281: instead of polluting `PeeredVPCTest` with knowledge about add-on tests, what do you think about wrapping it (like `AddOnEC2Test`) and add these helper methods in there?

test/e2e/suite/addons/addons_test.go
```diff
@@ -0,0 +1,110 @@
...
+		var test *suite.PeeredVPCTest
+
+		BeforeEach(func(ctx context.Context) {
+			test = suite.BeforeVPCTest(ctx, suiteConfig)
+		})
+
+		When("using ec2 instance as hybrid nodes", func() {
+			It("runs node monitoring agent tests", func(ctx context.Context) {
+				nodeMonitoringAgent := test.NewNodeMonitoringAgentTest()
+				test.Logger.Info("Running test for node monitoring agent")
```
g-gaston, line 90: nit: does this log provide much value?it tells you what the name of the test already says and it runs at the very beginning of it

test/e2e/kubernetes/verify.go
```diff
@@ -84,7 +84,7 @@ func (v VerifyNode) Run(ctx context.Context) error {
 	v.Logger.Info("Successfully exec'd nginx -version", "pod", podName)
 
 	v.Logger.Info("Checking logs for nginx output", "pod", podName)
-	logs, err := GetPodLogsWithRetries(ctx, v.K8s, podName, testPodNamespace)
+	logs, err := GetPodLogsWithRetries(ctx, v.K8s, podName, testPodNamespace, &corev1.PodLogOptions{})
```
g-gaston, line 87: not a blocker
The fact that you had to add the empty string to 3 calls already show that the customization of the logs calls is probably not the common usecase.

In general, I think making interfaces that simplify use for the most common usecase while allowing customization for the rest is a good rule of thumb

In this case, could we make it a variadic argument>

test/e2e/kubernetes/resource.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const (
+	resourceWaitTimeout  = 3 * time.Minute
+	resourcePollInterval = 5 * time.Second
+	maxConsecutiveErrors = 3
+)
+
+// WaitForResource is a generic function to wait until a kubernetes resource is ready.
+func WaitForResource(
```
g-gaston, line 19: I appreciate the intention here
However, are we gaining enough here to justify the complexity? It seems like the only reusable piece is the "error retry". But In order to do that, we need to define a pretty complex interface that takes 2 closures.

I wonder if maybe we should focus on that: have a reusable piece that retries errors for a determined amount of times. Then if you want to wait for resource ready, you wrap you code with a wait.Poll, which is already available everywhere.

test/e2e/addon/addon.go
```diff
@@ -83,3 +98,75 @@ func (a Addon) WaitUtilActive(ctx context.Context, client *eks.Client, logger lo
...
+	_, err := client.DeleteAddon(ctx, params)
+
+	if err == nil || errors.IsType(err, &types.ResourceNotFoundException{}) {
+		// Ignore if add-on doesn't exist
+		return nil
+	}
+	return err
+}
+
+func (a Addon) FetchLogs(ctx context.Context, k8s clientgo.Interface, logger logr.Logger) error {
```
g-gaston, line 131: imo this method does fit the addon abstraction. We are representing an eks add-on. This add-on can be created, deleted and has a status that can change overtime. So the other methods make sense.

But why fetch logs? I understand most add-ons deploy pods, but that's not part of the add-on concept. This is a bit academical, but I fear if we break the abstraction like this, we just open the doors for this to become a dumping ground.

It seems like the only things you need from the add-on resource is name and namespace. Why not just make that arguments to the method and free it from the add-on abstraction?

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,261 @@
...
+							SecurityContext: &v1.SecurityContext{
+								Privileged: ptr.To(true),
+								RunAsUser:  ptr.To(int64(0)),
+							},
+							Command: []string{
+								"/bin/bash",
+								"-c",
+								"echo 'watchdog: BUG: soft lockup - CPU#6 stuck for 23s! [VM Thread:4054]' | tee -a /host/dev/kmsg",
```
g-gaston, line 103: this is interesting
Why did you decide to do it with a daemon set vs just looping through the nodes and running a command on each?

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,261 @@
...
+func (n NodeMonitoringAgentTest) Create(ctx context.Context) error {
+	if err := n.addon.CreateAddon(ctx, n.EKSClient, n.K8S, n.Logger); err != nil {
+		return err
+	}
+
+	if err := kubernetes.WaitForDaemonSetReady(ctx, n.Logger, n.K8S, n.addon.Namespace, n.addon.Name); err != nil {
+		return err
+	}
+
+	if err := n.deployKernelError(ctx); err != nil {
```
g-gaston, line 69: nit: imo this should be part of `Validate`

test/e2e/addon/addon.go
tatlat: Moved it to `test/e2e/kubernetes/pod.go`

test/e2e/addon/nodemonitoringagent.go
tatlat: it didn't occur to me that I could run this on the node. I'll do that instead.

test/e2e/suite/peered_vpc.go
tatlat: i created `test/e2e/suite/addon_ec2.go` and moved it there

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,197 @@
...
+		return err
+	}
+
+	timesRun := 0
+	for _, node := range nodes.Items {
+		for _, ip := range node.Status.Addresses {
+			if ip.Type != v1.NodeInternalIP {
+				continue
+			}
+			if _, err := n.CommandRunner.Run(ctx, ip.Address, []string{"echo 'watchdog: BUG: soft lockup - CPU#6 stuck for 23s! [VM Thread:4054]' | tee -a /dev/kmsg"}); err != nil {
```
g-gaston, line 86: nit: comment here explaining what this command does and why we use it?

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,197 @@
...
+	nodes, err := ik8s.ListRetry(ctx, n.K8S.CoreV1().Nodes())
+	if err != nil {
+		return err
+	}
+
+	timesRun := 0
+	for _, node := range nodes.Items {
+		for _, ip := range node.Status.Addresses {
+			if ip.Type != v1.NodeInternalIP {
+				continue
```
g-gaston, line 84: why not just return error here? you end up returning an error if count doesn't match number of nodes, but because you hide the cause, the error doesn't say why it didn't run for all nodes

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,197 @@
...
+	return nil
+}
+
+func (n *NodeMonitoringAgentTest) Validate(ctx context.Context) error {
+	nodes, err := ik8s.ListRetry(ctx, n.K8S.CoreV1().Nodes())
+	if err != nil {
+		return err
+	}
+
+	if err := n.runKernelError(ctx); err != nil {
```
g-gaston, line 106: why not pass the list of nodes to `runKernelError` if you already have it here?

test/e2e/kubernetes/pod.go
```diff
@@ -154,6 +152,31 @@ func DeletePod(ctx context.Context, k8s kubernetes.Interface, name, namespace st
 	return waitForPodToBeDeleted(ctx, k8s, name, namespace)
 }
 
+// FetchLogs retrieves logs for all pods with the name prefix and returns it in a single string.
+func FetchLogs(ctx context.Context, k8s kubernetes.Interface, name, namespace string) (string, error) {
+	var pods *corev1.PodList
+
+	listOpts := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", "app.kubernetes.io/name", name),
```
g-gaston, line 160: this seems super specific to your usecase
why don't you pass kubernetes.ListOptions to this method? you can set the label selector in the caller

test/e2e/suite/addons/addons_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// a struct that we can make accessible from the tests. We leave the rest
+	// for the per tests setup code.
+	func(ctx context.Context, data []byte) {
+		suiteConfig = suite.BeforeSuiteCredentialUnmarshal(ctx, data)
+	},
+)
+
+var _ = Describe("Hybrid Nodes", func() {
+	When("using peered VPC", func() {
+		var test *suite.PeeredVPCTest
```
g-gaston, line 81: why do we need this variable?

test/e2e/suite/addons/addons_test.go
```diff
@@ -0,0 +1,111 @@
...
+					report := CurrentSpecReport()
+					if report.State.Is(types.SpecStateFailed) {
+						err := nodeMonitoringAgent.CollectLogs(ctx)
+						if err != nil {
+							// continue cleanup even if logs collection fails
+							GinkgoWriter.Printf("Failed to collect node monitoring agent logs: %v\n", err)
+						}
+					}
+					Expect(nodeMonitoringAgent.Delete(ctx)).To(Succeed(), "should cleanup node monitoring agent successfully")
+				})
```
g-gaston, line 103: not a blocker
you run this even if the addon fails to be created. what logs will you get in that case?
this is one of the issues of abstracting things too early in the Run method: you don't have control in the test code of the different steps
if so, you could define the defers at the right point: delete after calling create if create succeeds, logs if pods don't deploy, etc.

test/e2e/addon/nodemonitoringagent.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	return slices.Contains(monitoringConditionTypes, string(condition.Type))
+}
+
+func isMonitoringAgentEvent(event v1.Event) bool {
+	return strings.Contains(strings.ToLower(event.Source.Component), nodeMonitoringAgentName)
+}
+
+func (n *NodeMonitoringAgentTest) CollectLogs(ctx context.Context) error {
```
g-gaston, line 164: this is more like
```suggestion
func (n *NodeMonitoringAgentTest) PrintLogs(ctx context.Context) error {
```

test/e2e/suite/addons/addons_test.go
```diff
@@ -0,0 +1,117 @@
...
+	// for the per tests setup code.
+	func(ctx context.Context, data []byte) {
+		suiteConfig = suite.BeforeSuiteCredentialUnmarshal(ctx, data)
+	},
+)
+
+var _ = Describe("Hybrid Nodes", func() {
+	When("using peered VPC", func() {
+		var addonEc2Test *suite.AddonEc2Test
+		var createSucceeded bool
```
g-gaston, line 82: you can define this in the It scope, it doesn't need to be up here. this will be shared by other It blocks, and i don't think that's necessary

test/e2e/suite/addons/addons_test.go
```diff
@@ -0,0 +1,117 @@
...
+					report := CurrentSpecReport()
+					if report.State.Is(types.SpecStateFailed) && createSucceeded {
+						err := nodeMonitoringAgent.CollectLogs(ctx)
+						if err != nil {
+							// continue cleanup even if logs collection fails
+							GinkgoWriter.Printf("Failed to collect node monitoring agent logs: %v\n", err)
+						}
+					}
+					Expect(nodeMonitoringAgent.Delete(ctx)).To(Succeed(), "should cleanup node monitoring agent successfully")
+				})
```
g-gaston, line 103: why don't you divide this in two defers?
one for the Delete and one for the logs
you can leave the delete up here
and define the logs one only after the create is succesful

g-gaston (approved):

PR #522, opened 2025-04-30
Bump pod identity timeout to 10 mins

achen1210:
*Issue #, if available:*

*Description of changes:*

Bump timeout to 10 mins for pod identity add on since we're seeing lots of context exceeded errors due to 2 minute timeout; from logs it has looked like it's taken up to 7 mins to add on to be ready (often due to waiting on a single pod in daemonset to be in ready state)

*Testing (if applicable):*

### DIFF
--- a/test/e2e/addon/podidentityverify.go
+++ b/test/e2e/addon/podidentityverify.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	getAddonTimeout           = 2 * time.Minute
+	getAddonTimeout           = 8 * time.Minute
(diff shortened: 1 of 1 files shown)


test/e2e/addon/podidentityverify.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	getAddonTimeout           = 2 * time.Minute
+	getAddonTimeout           = 8 * time.Minute
```
junpengdev, line 23: Let's bump it to 10. We already see one taking 7 min to become ready. Changing it to 8 min seems still a little small.

junpengdev (approved):

PR #528, opened 2025-05-05
Remove pods with uninstall

vignesh-goutham:
*Description of changes:*
On removing containerd, the pods are not explicitly deleted. This is by design and containerd lets cri plugin or rather kubelet be the handler of these processes. When containerd exists, the pod sandboxes, containers, process are not terminated. Kubelet currently while attempts to stop pods, doesn't force clean up sandboxes. This leads to nodes to have pods running after nodeadm uninstall. This change finds and deletes running pod sandboxes and containers.

Tested manually.

### DIFF
--- a/internal/containerd/install.go
+++ b/internal/containerd/install.go
@@ -7,12 +7,16 @@ import (
 	"os/exec"
 	"time"
 
+	"github.com/containerd/containerd/integration/remote"
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	v1 "k8s.io/cri-api/pkg/apis/runtime/v1"
 
 	"github.com/aws/eks-hybrid/internal/artifact"
 	"github.com/aws/eks-hybrid/internal/daemon"
 	"github.com/aws/eks-hybrid/internal/system"
 	"github.com/aws/eks-hybrid/internal/tracker"
+	"github.com/aws/eks-hybrid/internal/util"
 	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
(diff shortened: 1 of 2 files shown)


internal/containerd/install.go
```diff
@@ -127,3 +132,75 @@ func isContainerdNotInstalled() bool {
...
+}
+
+func (c *Client) RemovePods() error {
+	podSandboxes, err := c.Runtime.ListPodSandbox(&v1.PodSandboxFilter{
+		State: &v1.PodSandboxStateValue{
+			State: v1.PodSandboxState_SANDBOX_READY,
+		},
+	})
+	if err != nil {
+		return err
```
jaxesn, line 158: lets wrap/fmt/whatever these errs with a custom message to make tracking later easier.

internal/containerd/install.go
```diff
@@ -127,3 +132,75 @@ func isContainerdNotInstalled() bool {
...
+				return err
+			}
+			return nil
+		})
+		if err != nil {
+			zap.L().Info("ignored error stopping pod", zap.Error(err))
+		}
+	}
+
+	// If pod sandbox deletes dont work, we can try to stop and remove containers individually
```
jaxesn, line 177: `// If pod sandbox deletion fails, we can try to stop and remove containers individually`

internal/containerd/install.go
```diff
@@ -127,3 +132,75 @@ func isContainerdNotInstalled() bool {
...
+	containers, err := c.Runtime.ListContainers(nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to list containers")
```
jaxesn, line 180: errors.Wrap(err, "listing containers")

internal/containerd/install.go
jaxesn: looks like we are going to try either way?  im assuming if sandboxes cleanly delete, then the listcontainers will return an empty list?

internal/containerd/install.go
```diff
@@ -127,3 +132,75 @@ func isContainerdNotInstalled() bool {
...
+func NewClient() (*Client, error) {
+	runtime, err := remote.NewRuntimeService(ContainerRuntimeEndpoint, 5*time.Second)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{Runtime: runtime}, nil
+}
+
+func (c *Client) RemovePods() error {
+	podSandboxes, err := c.Runtime.ListPodSandbox(&v1.PodSandboxFilter{
```
jaxesn, line 152: for these containerd calls, do we need to add the namespace somewhere, or does it default to the k8s.io one?

internal/containerd/install.go
```diff
@@ -127,3 +132,75 @@ func isContainerdNotInstalled() bool {
...
+	for _, container := range containers {
+		status, err := c.Runtime.ContainerStatus(container.Id)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get container status for %s", container.Id)
+		}
+		zap.L().Info("Stopping container..", zap.String("container", container.Metadata.Name))
+		err = util.RetryExponentialBackoff(3, 2*time.Second, func() error {
+			if status.State == v1.ContainerState_CONTAINER_RUNNING {
+				if err := c.Runtime.StopContainer(container.Id, 0); err != nil {
+					return errors.Wrapf(err, "failed to stop container %s", container.Id)
```
jaxesn, line 192: `return errors.Wrapf(err, "stopping container %s", container.Id)`

internal/containerd/install.go
```diff
@@ -127,3 +132,75 @@ func isContainerdNotInstalled() bool {
...
+			}
+			return nil
+		})
+		if err != nil {
+			zap.L().Info("ignored error stopping pod", zap.Error(err))
+		}
+	}
+
+	// If pod sandbox deletes dont work, we can try to stop and remove containers individually
+	containers, err := c.Runtime.ListContainers(nil)
```
jaxesn, line 178: what is the var we are pasing nil for?

jaxesn (changes requested):

internal/containerd/install.go
vignesh-goutham: Thats right, if sandboxes go away, listcontainers returns empty list.

internal/containerd/install.go
vignesh-goutham: This actually uses the cri-api, which only works with the k8s.io namespace.

internal/containerd/install.go
vignesh-goutham: This is a filter for list containers call. We also want to remove stopped containers here, since post containerd daemon stop, stopped containers are not anymore GCed.

PR #532, opened 2025-05-06
Add retries around kubernetes validations

ahreehong:
*Issue #, if available:*

*Description of changes:*
Seeing issues in the e2e tests where the validations in the debug command fails due to timeouts. Wrapping these operations to retry to fix the flakiness

*Testing (if applicable):*

### DIFF
--- a/internal/iamrolesanywhere/validate.go
+++ b/internal/iamrolesanywhere/validate.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/aws/eks-hybrid/internal/api"
 	"github.com/aws/eks-hybrid/internal/network"
+	"github.com/aws/eks-hybrid/internal/retry"
 	"github.com/aws/eks-hybrid/internal/validation"
 )
 
@@ -24,7 +25,10 @@ func CheckEndpointAccess(ctx context.Context, config aws.Config) error {
 		return fmt.Errorf("resolving IAM Roles Anywhere endpoint: %w", err)
 	}
 
-	if err := network.CheckConnectionToHost(ctx, endpoint.URI); err != nil {
+	err = retry.NetworkRequest(ctx, func(ctx context.Context) error {
+		return network.CheckConnectionToHost(ctx, endpoint.URI)
+	})
+	if err != nil {
 		return fmt.Errorf("checking connection to IAM Roles Anywhere endpoint: %w", err)
 	}
 
--- a/internal/kubernetes/connection.go
+++ b/internal/kubernetes/connection.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/eks-hybrid/internal/api"
 	"github.com/aws/eks-hybrid/internal/network"
+	"github.com/aws/eks-hybrid/internal/retry"
 	"github.com/aws/eks-hybrid/internal/validation"
 )
 
@@ -23,7 +24,10 @@ func CheckConnection(ctx context.Context, informer validation.Informer, node *ap
 		return err
 	}
 
-	if err = network.CheckConnectionToHost(ctx, *endpoint); err != nil {
+	err = retry.NetworkRequest(ctx, func(ctx context.Context) error {
+		return network.CheckConnectionToHost(ctx, *endpoint)
+	})
+	if err != nil {
 		err = validation.WithRemediation(err, "Ensure your network configuration allows the node to access the Kubernetes API endpoint.")
 		return err
 	}
--- a/internal/kubernetes/unauthenticated.go
+++ b/internal/kubernetes/unauthenticated.go
@@ -5,12 +5,14 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
 
+	"github.com/pkg/errors"
+
 	"github.com/aws/eks-hybrid/internal/api"
+	"github.com/aws/eks-hybrid/internal/retry"
 	"github.com/aws/eks-hybrid/internal/validation"
 )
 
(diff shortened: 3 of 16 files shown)


internal/node/validate.go
```diff
@@ -52,7 +53,18 @@ func (a APIServerValidator) MakeAuthenticatedRequest(ctx context.Context, inform
 		return err
 	}
 
-	_, err = client.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, validation.ValidationInterval, validation.ValidationTimeout, true, func(ctx context.Context) (bool, error) {
+		_, err = client.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
```
jaxesn, line 58: we do the endpoints call in this file on  line 143 as well. we should cover that case too.

internal/node/validate.go
```diff
@@ -140,7 +153,19 @@ func (a APIServerValidator) CheckVPCEndpointAccess(ctx context.Context, informer
 		return err
 	}
 
-	kubeEndpoint, err := client.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
+	var kubeEndpoint *v1.Endpoints
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, validation.ValidationInterval, validation.ValidationTimeout, true, func(ctx context.Context) (bool, error) {
```
g-gaston, line 158: are these all the requests we make from debug? i suspect there are more. at least we have one that calls the selfsubjectreview api, there might be more

internal/node/validate.go
g-gaston, line 1: In line 200 you have a call to `network.CheckConnectionToHost` in a loop. why wouldn't we retry that?

internal/kubernetes/unauthenticated.go
```diff
@@ -35,7 +37,19 @@ func MakeUnauthenticatedRequest(ctx context.Context, endpoint string, caCertific
 		return validation.WithRemediation(err, "Ensure the Kubernetes API server endpoint provided is correct.")
 	}
 
-	resp, err := client.Do(req)
+	consecutiveErrors := 0
+	var resp *http.Response
+	err = wait.PollUntilContextTimeout(ctx, validation.ValidationInterval, validation.ValidationTimeout, true, func(ctx context.Context) (bool, error) {
```
g-gaston, line 42: you might want to extend this loop to line 59. what if you dial in correctly but fail reading the rest of the response? the network connection can be interrupted at any point of the request

internal/kubernetes/connection.go
```diff
@@ -23,7 +25,19 @@ func CheckConnection(ctx context.Context, informer validation.Informer, node *ap
 		return err
 	}
 
-	if err = network.CheckConnectionToHost(ctx, *endpoint); err != nil {
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, validation.ValidationInterval, validation.ValidationTimeout, true, func(ctx context.Context) (bool, error) {
+		err = network.CheckConnectionToHost(ctx, *endpoint)
```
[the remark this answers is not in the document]
ahreehong: Nope, I just missed those because i didn't see them error. I took a pass through all of the requests we make in debug and added retries

internal/kubernetes/connection.go
```diff
@@ -23,7 +25,19 @@ func CheckConnection(ctx context.Context, informer validation.Informer, node *ap
 		return err
 	}
 
-	if err = network.CheckConnectionToHost(ctx, *endpoint); err != nil {
+	consecutiveErrors := 0
+	err = wait.PollUntilContextTimeout(ctx, validation.ValidationInterval, validation.ValidationTimeout, true, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
ahreehong: added a standardized re-trier here: https://github.com/aws/eks-hybrid/pull/550

internal/iamrolesanywhere/validate.go
```diff
@@ -6,25 +6,36 @@ import (
...
-	endpoint, err := opts.EndpointResolverV2.ResolveEndpoint(ctx, rolesanywhere.EndpointParameters{
-		Region:   aws.String(opts.Region),
-		Endpoint: opts.BaseEndpoint,
+	var endpoint transport.Endpoint
+	err := retry.NetworkRequest(ctx, func(ctx context.Context) error {
+		var err error
+		endpoint, err = opts.EndpointResolverV2.ResolveEndpoint(ctx, rolesanywhere.EndpointParameters{
+			Region:   aws.String(opts.Region),
+			Endpoint: opts.BaseEndpoint,
+		})
```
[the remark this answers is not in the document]
ahreehong: Whoops, just double checked and it doesn't - I think I got it mixed up with something else. I'll remove the retries from here

internal/kubernetes/access.go
```diff
@@ -23,7 +24,12 @@ func NewAccessValidator(config aws.Config) AccessValidator {
 func (a AccessValidator) Run(ctx context.Context, informer validation.Informer, node *api.NodeConfig) error {
 	// APIServerEndpoint, CertificateAuthority are required for the validations we want to run here but are optional in the config
 	// When not specified, we need to read them from the EKS API.
-	cluster, err := eks.ReadClusterDetails(ctx, a.aws, node)
+	var cluster *api.ClusterDetails
+	err := retry.NetworkRequest(ctx, func(ctx context.Context) error {
+		var err error
+		cluster, err = eks.ReadClusterDetails(ctx, a.aws, node)
```
g-gaston, line 30: I'm not sure we need this here. The aws client already handles retries
@jaxesn

jaxesn: Yeah we can remove this. We use the defaults on the aws config, which are up to 3 retries, which I think is fine for nodeadm commands (in the e2e tests we bump it to like 20 since we are making so many concurrent requests).  If we do see issues tho we can tweak the retry config on the aws config.

jaxesn (approved):

Issue #534, opened 2025-05-08
upgrading on rhel based OS fails to update containerd


jaxesn:
**What happened**:

When running `nodeadm upgrade` on a rhel based instance we are seeing:

```
{"level":"info","ts":1746714597.9160807,"caller":"upgrade/upgrade.go:87","msg":"Loading installed components"}
{"level":"info","ts":1746714597.9162924,"caller":"upgrade/upgrade.go:100","msg":"Validating if node has initialized"}
{"level":"info","ts":1746714598.1217346,"caller":"upgrade/upgrade.go:106","msg":"Loading configuration..","configSource":"file://eks-config.yaml/"}
{"level":"info","ts":1746714598.1218228,"caller":"node/node.go:13","msg":"Loading configuration..","configSource":"file://eks-config.yaml/"}
{"level":"info","ts":1746714598.1223435,"caller":"node/node.go:23","msg":"Setting up hybrid node provider..."}
{"level":"info","ts":1746714598.1231966,"caller":"hybrid/validator.go:60","msg":"Validating configuration..."}
{"level":"info","ts":1746714598.1232255,"caller":"upgrade/upgrade.go:132","msg":"Validating Kubernetes version","kubernetes version":"1.31"}
{"level":"info","ts":1746714598.2608845,"caller":"upgrade/upgrade.go:138","msg":"Using Kubernetes version","kubernetes version":"1.31.7"}
{"level":"info","ts":1746714598.2609313,"caller":"upgrade/upgrade.go:140","msg":"Creating daemon manager.."}
{"level":"info","ts":1746714598.2619846,"caller":"upgrade/upgrade.go:154","msg":"Validating if node has been drained..."}
{"level":"info","ts":1746714598.3443599,"caller":"upgrade/upgrade.go:163","msg":"Validating if node has been marked unschedulable..."}
{"level":"info","ts":1746714598.387244,"caller":"upgrade/upgrade.go:171","msg":"Creating package manager..."}
{"level":"info","ts":1746714598.387265,"caller":"upgrade/upgrade.go:173","msg":"Configuring package manager with","containerd source":"none"}
{"level":"info","ts":1746714598.3873127,"caller":"flows/upgrade.go:65","msg":"Refreshing package manager metadata cache..."}
{"level":"info","ts":1746714602.8865938,"caller":"flows/upgrade.go:70","msg":"Upgrading containerd..."}
[...]

PR #535, opened 2025-05-09
fixes upgrade and uninstall when using containerd-source:none

jaxesn:
*Issue #, if available:*

*Description of changes:*
see #534 for background

Fixes handling of containerd-source:none during upgrade and install.  Previously during upgrade nodeadm was yum/apt updating and during uninstall it would have removed containerd even if the user had ran install with source:none.

This PR is broken into 3 commits:
- the first commit adds an integration test to prove the failure, replicating the error on al23 requires a bit of yum magic since containerd is provided in the yum repo, whereas on rhel yum gets confused easier since there is no containerd package in the default repos
- the second commit is the bare minimum fix for this issue where because we were not properly handling the empty value of the Containerd "enum" in the tracker struct/file, we were comparing "" against None, which was false and therefore dropping into the upgrade/uninstall flow
- the third commit improves the handling of this default value by ensuring we always write and read empty string as `none`, thus avoiding this class of error in the future.

*Testing (if applicable):*

### DIFF
--- a/cmd/nodeadm/install/install.go
+++ b/cmd/nodeadm/install/install.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/eks-hybrid/internal/logger"
 	"github.com/aws/eks-hybrid/internal/packagemanager"
 	"github.com/aws/eks-hybrid/internal/ssm"
+	"github.com/aws/eks-hybrid/internal/tracker"
 )
 
 const installHelpText = `Examples:
@@ -29,7 +30,8 @@ Documentation:
 
 func NewCommand() cli.Command {
 	cmd := command{
-		timeout: 20 * time.Minute,
+		timeout:          20 * time.Minute,
+		containerdSource: string(tracker.ContainerdSourceDistro),
 	}
 	cmd.region = ssm.DefaultSsmInstallerRegion
 
@@ -79,11 +81,10 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 		return err
 	}
 
-	// Default containerd source to distro
-	if c.containerdSource == "" {
-		c.containerdSource = string(containerd.ContainerdSourceDistro)
+	containerdSource, err := tracker.ContainerdSource(c.containerdSource)
+	if err != nil {
+		return err
 	}
-	containerdSource := containerd.GetContainerdSource(c.containerdSource)
 	if err := containerd.ValidateContainerdSource(containerdSource); err != nil {
 		return err
 	}
--- a/internal/containerd/install.go
+++ b/internal/containerd/install.go
@@ -16,12 +16,7 @@ import (
 	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
-type SourceName string
-
 const (
-	ContainerdSourceNone   SourceName = "none"
-	ContainerdSourceDistro SourceName = "distro"
-	ContainerdSourceDocker SourceName = "docker"
 	// pin containerd to major version 1.x
 	ContainerdVersion = "1.*"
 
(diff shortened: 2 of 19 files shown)


cmd/nodeadm/install/install.go
```diff
@@ -81,9 +82,9 @@ func (c *command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 
 	// Default containerd source to distro
 	if c.containerdSource == "" {
-		c.containerdSource = string(containerd.ContainerdSourceDistro)
+		c.containerdSource = string(tracker.ContainerdSourceDistro)
 	}
```
g-gaston, line 86: imo the right way to do this is to define the default in the `cmd` variable in line 32. That way it shows up in the command help

I know this is not your code though

internal/containerd/install.go
```diff
@@ -34,20 +29,19 @@ type Source interface {
 	GetContainerd(version string) artifact.Package
 }
 
-func Install(ctx context.Context, tracker *tracker.Tracker, source Source, containerdSource SourceName) error {
-	if containerdSource == ContainerdSourceNone {
+func Install(ctx context.Context, artifactsTracker *tracker.Tracker, source Source, containerdSource tracker.SourceName) error {
+	if containerdSource == tracker.ContainerdSourceNone || areContainerdAndRuncInstalled() {
```
g-gaston, line 33: This is where i got lost the other day. help me understand in what case the condition for `areContainerdAndRuncInstalled` is useful: why would we skip installation if the user is asking as to install but they have something already installed? In my head we would just error if the user already has it but they are not using option "none"


or is this for the upgrade flow where we need to deal with a tracker file with an empty `containerdSource`?

internal/containerd/install.go
```diff
@@ -34,20 +29,19 @@ type Source interface {
...
-		if err := cmd.Retry(ctx, containerd.InstallCmd, 5*time.Second); err != nil {
-			return errors.Wrap(err, "failed to install containerd")
-		}
-		tracker.MarkContainerd(string(containerdSource))
+	containerd := source.GetContainerd(ContainerdVersion)
+	// Sometimes install fails due to conflicts with other processes
+	// updating packages, specially when automating at machine startup.
+	// We assume errors are transient and just retry for a bit.
+	if err := cmd.Retry(ctx, containerd.InstallCmd, 5*time.Second); err != nil {
+		return errors.Wrap(err, "failed to install containerd")
```
g-gaston, line 42: nit
```suggestion
		return errors.Wrap(err, "installing containerd")
```

internal/tracker/artifacts.go
```diff
@@ -56,12 +64,10 @@ func (tracker *Tracker) Add(componentName string) error {
...
 }
 
-func (tracker *Tracker) MarkContainerd(source string) {
-	tracker.Artifacts.Containerd = source
-}
-
 // Save() saves the tracker to file
 func (tracker *Tracker) Save() error {
+	// ensure containerd source is populated with none/distro/docker
+	tracker.Artifacts.Containerd = GetContainerdSource(string(tracker.Artifacts.Containerd))
```
g-gaston, line 70: this makes me a bit nervous
if someone adds a new value to the enum but forgets to update `GetContainerdSource`, this will all silently work but default to "none"

What would you think about handling "" and "none" as `ContainerdSourceNone`, but default to error in `GetContainerdSource`? that way we enforce only valid values can be written

I'll admit this is a bit paranoid but you know

internal/tracker/artifacts.go
```diff
@@ -103,3 +113,14 @@ func GetCurrentState() (*Tracker, error) {
 	}
 	return tracker, nil
 }
+
+func GetContainerdSource(containerdSource string) SourceName {
```
g-gaston, line 117: nit
```suggestion
func ContainerdSource(containerdSource string) SourceName {
```

internal/tracker/artifacts.go
```diff
@@ -13,14 +13,22 @@ import (
 	"github.com/aws/eks-hybrid/internal/util"
 )
 
+type SourceName string
```
g-gaston, line 16:
```suggestion
type ContainerdSourceName string
```

g-gaston (approved):

PR #543, opened 2025-05-13
implements untar/gz in go intead of exec-ing out

jaxesn:
*Issue #, if available:*

*Description of changes:*

Use golang std gzip/tar libs to extract the cni plugins tgz instead of exec-ing to tar.

*Testing (if applicable):*

### DIFF
--- a/internal/artifact/install_test.go
+++ b/internal/artifact/install_test.go
@@ -1,11 +1,15 @@
 package artifact_test
 
 import (
+	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"testing"
+
+	. "github.com/onsi/gomega"
 
 	"github.com/aws/eks-hybrid/internal/artifact"
 )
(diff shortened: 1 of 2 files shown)


internal/artifact/install_test.go
```diff
@@ -82,3 +86,243 @@ func TestInstallFile_DirNotExists(t *testing.T) {
...
+		name  string
+		files map[string]struct {
+			content string
+			mode    int64
+		}
+		setup  func(g *GomegaWithT, dst string)
+		verify func(g *GomegaWithT, dst string, files map[string]struct {
+			content string
+			mode    int64
+		})
```
g-gaston, line 147: not a blocker: imo this pattern tends to make tests more difficult to read. i understand the goal is to be able to fit all cases in one single table test case, but i think sometimes test cases are simply too far apart.

What I usually do is keep the table test for all the cases i can find that follow a simple "in -> out" pattern. And for anything that either needs special setup or validation, then write separate tests. To avoid duplication, just create helper methods for the share test logic. This is just a rule of thumb, sometimes the test complexity is worth it to keep everything together.

Not asking to change this, I don't anticipate this code will change a lot so the tests will probably be fairly stable, so maintainability concerns are not that bad. Just sharing my thoughts on the pattern since I noticed it's becoming more common lately.

g-gaston (approved):

PR #544, opened 2025-05-15
adds support for gzipped uri in artifacts manifest

jaxesn:
*Issue #, if available:*

*Description of changes:*

We are going to start publish .gz files for the individual binaries we download via nodeadm install.  This adds support for these in the manifest, which does not currently exist today, but will after the next platform release.  Currently during install we download ~200MB from the cloudfront distro, once these are all available as gz versions this will drop down to ~50MB.

*Testing (if applicable):*

### DIFF
--- a/internal/artifact/source.go
+++ b/internal/artifact/source.go
@@ -1,6 +1,7 @@
 package artifact
 
 import (
+	"compress/gzip"
 	"fmt"
 	"hash"
 	"io"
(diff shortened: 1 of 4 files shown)


internal/artifact/source.go
```diff
@@ -56,3 +71,13 @@ func WithNopChecksum(rc io.ReadCloser) Source {
 		ChecksumVerifier: nopChecksumVerifier{},
 	}
 }
+
+type gzipReadCloser struct {
+	*gzip.Reader
```
g-gaston, line 76: nit: i would not embed this one, it makes the implementation of Close() a bit less clear imo

g-gaston (approved):

PR #551, opened 2025-05-19
updates pod/node retries in e2e

jaxesn:
*Issue #, if available:*

*Description of changes:*

Switch most of the kubernetes retries in e2e to use the new helpers.  Most changes are pretty straightforward.  

For a couple, like WaitForNode and WaitForPod variants, I used the ListAndWait with a listOption filter vs using GetAndWait which does not give the error we would need to check if it was NotFound.

There are a few remaining uses of retry.OnError I have left for a future PR (if at all).  These are mostly cases that may be longer running actions, such as drain or getPodLogs.

*Testing (if applicable):*

### DIFF
--- a/internal/kubernetes/retry.go
+++ b/internal/kubernetes/retry.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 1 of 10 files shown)


test/e2e/kubernetes/pod.go
```diff
@@ -92,64 +91,65 @@ func CreatePod(ctx context.Context, k8s kubernetes.Interface, pod *corev1.Pod, l
...
-				return false, nil // continue polling
+				return false
 			}
 		}
-
-		return true, nil // pod is running, stop polling
+		// stop polling
+		return true
+	}, func(opts *ik8s.ListOptions) {
+		opts.FieldSelector = listOptions.FieldSelector
```
g-gaston, line 119: maybe?
```suggestion
		opts.ListOptions = listOptions
```

internal/kubernetes/wait.go
```diff
@@ -10,7 +10,10 @@ import (
 	"github.com/aws/eks-hybrid/internal/retry"
 )
 
-const minWait = time.Duration(200 * time.Millisecond)
+const (
+	minWait = time.Duration(200 * time.Millisecond)
```
g-gaston, line 14: nit
this was "wrong" before
```suggestion
	minWait = 200 * time.Millisecond
```

internal/kubernetes/retry.go
```diff
@@ -72,6 +73,83 @@ func ListRetry[O runtime.Object](ctx context.Context, lister Lister[O], opts ...
...
+	err := retryRequest(ctx, func(ctx context.Context) error {
+		err := deleter.Delete(ctx, name, deleteOpt.DeleteOptions)
+		if apierrors.IsNotFound(err) {
+			notFoundErr = err
+			return nil
+		}
+		return err
+	})
+	if notFoundErr != nil {
+		return notFoundErr
```
g-gaston, line 108: not totally against this, but let's imagine this scenario:
1. We make first delete request. the object exists, the api server deletes it correctly from etcd and returns the 200 response, but the connection gets interrupted and the tcp session is aborted before we get the response.
2. We see this as a i/o timeout or connection reset by peer on our side. So we retry.
3. The API server receives the new request but the object has already been deleted, so it returns not found error.
4. We return the not found error as is to our caller.

From the caller perspective this is a bit confusing: does this mean the object didn't exist? 

Maybe it's because I'm assuming the same semantics as the other retry operations and we just need to define the semantics of retries for write operations. What do we want to retry for these? only i/o or connection errors? In that case maybe we could just check if the error is an API error and only retry if it's not. In that case the semantics would be: we retry until we get a response from the API server.

Not trying to be pedantic but I feel the semantics we want here are different than for the read operations. I find valuable to define them more concretely so the expectations are clear and callers can build accordingly.

internal/kubernetes/retry.go
```diff
@@ -72,6 +73,83 @@ func ListRetry[O runtime.Object](ctx context.Context, lister Lister[O], opts ...
...
+		var err error
+		created, err = creator.Create(ctx, obj, createOpt.CreateOptions)
+		if apierrors.IsAlreadyExists(err) {
+			alreadyExistsErr = err
+			return nil
+		}
+		return err
+	})
+	if alreadyExistsErr != nil {
+		return created, alreadyExistsErr
```
g-gaston, line 147: interesting
will created have the current value of the existing object?

test/e2e/kubernetes/serviceaccount.go
```diff
@@ -2,40 +2,31 @@ package kubernetes
...
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
 
-		if _, err := k8s.CoreV1().ServiceAccounts(namespace).Create(ctx, serviceAccount, metav1.CreateOptions{}); err != nil {
-			return err
-		}
+	err := ik8s.IdempotentCreate(ctx, k8s.CoreV1().ServiceAccounts(namespace), sa)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
```
g-gaston, line 27: nit
```suggestion
	if err != nil {
```

g-gaston (approved):

PR #555, opened 2025-05-20
Add retries for ssm agent registration

ppadyala1599:
Adding retries with backoff for SSM agent registration to handle network connectivity issues (connection reset errors). Uses cmd.Retry with a 10seconds backoff and 60seconds timeout allowing 6 retry attempts before failure

### DIFF
--- a/internal/ssm/config.go
+++ b/internal/ssm/config.go
@@ -13,12 +13,20 @@ import (
 	"github.com/aws/eks-hybrid/internal/util/cmd"
 )
 
+const (
+	// SSMRegistrationTimeout is the maximum time to wait for SSM registration to complete
+	SSMRegistrationTimeout = 60 * time.Second
+
+	// SSMRegistrationBackoff is the time to wait between registration retry attempts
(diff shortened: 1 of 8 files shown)


internal/ssm/config.go
```diff
@@ -31,16 +34,21 @@ func (s *ssm) registerMachine(cfg *api.NodeConfig) error {
...
+		cmdBuilder := func(ctx context.Context) *exec.Cmd {
+			return exec.CommandContext(ctx, agentPath,
+				"-register", "-y",
+				"-region", cfg.Spec.Cluster.Region,
+				"-code", cfg.Spec.Hybrid.SSM.ActivationCode,
+				"-id", cfg.Spec.Hybrid.SSM.ActivationID,
+			)
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
jaxesn, line 47: we need to pass the context to reigsterMachine so we can avoid calling Background() here

jaxesn (approved):

PR #577, opened 2025-05-30
Make e2e-test commands (setup, cleanup, node create, node delete) callable from other packages

junpengdev:
*Issue #, if available:*
Currently our e2e-test commands are only available from command line. This makes it harder to be called from other packages.

*Description of changes:*
In this PR, I am creating `NewXXXCommandWithParams` method for `setup`, `cleanup`, `node create`, `node delete` commands that other packages can import them to create and delete hybrid nodes.

*Testing (if applicable):*
Tested with aws-k8s-deployer

### DIFF
--- a/cmd/e2e-test/cleanup/cleanup.go
+++ b/cmd/e2e-test/cleanup/cleanup.go
@@ -34,6 +34,13 @@ func NewCommand() *Command {
 	return &cmd
 }
 
+func NewCommandWithParams(resourcesFilePath string) *Command {
+	cmd := Command{
+		resourcesFilePath: resourcesFilePath,
+	}
+	return &cmd
+}
+
 func (c *Command) Flaggy() *flaggy.Subcommand {
 	return c.flaggy
 }
--- a/cmd/e2e-test/node/create.go
+++ b/cmd/e2e-test/node/create.go
@@ -59,6 +59,20 @@ func NewCreateCommand() cli.Command {
 	return &cmd
 }
 
+func NewCreateCommandWithParams(name, configFile, credsProvider, osType, arch, instanceSize, instanceType string) cli.Command {
+	cmd := create{
+		os:            osType,
+		arch:          arch,
+		instanceSize:  instanceSize,
+		instanceType:  instanceType,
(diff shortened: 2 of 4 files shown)


jaxesn:
Im not sure this is the pattern we should use. Most (maybe all?) of these commands already use some package, can we not just use the package directly from the other locations?  If we do expose it via another package and need to move some of the code form the command into a different package, they may be a better approach.  The run-e2e is one example that we use from the canary code, maybe we can follow a similar approach?

junpengdev:
> Im not sure this is the pattern we should use. Most (maybe all?) of these commands already use some package, can we not just use the package directly from the other locations? If we do expose it via another package and need to move some of the code form the command into a different package, they may be a better approach. The run-e2e is one example that we use from the canary code, maybe we can follow a similar approach?

When you mention using package directly from other locations, which packages are you referring to? Are you suggesting to combine these commands in `eks-hybrid` repo then make it exportable to our packages?

jaxesn:
No, im suggesting just using for example, https://github.com/aws/eks-hybrid/pull/577/files#diff-f8dc987f8de8dd27ed2cfc8c54f224d1f8e1a011a11186e25ce0c5d622d70cc7R76, cluster.NewDelete/delete.Run directly from the other projects. I believe we have a similiar pattern for most/all of these commands where the command go is really just the wrapper which sets up logging/ctx/aws/etc which should likely be in the consumer project anyway

cmd/e2e-test/cleanup/cleanup.go
```diff
@@ -34,6 +34,13 @@ func NewCommand() *Command {
 	return &cmd
 }
 
+func NewCommandWithParams(resourcesFilePath string) *Command {
```
g-gaston, line 37: Could you explain why we need this?
In your PR description you indicate that the current code "makes it harder to be called from other packages". But, aren't our canary tests using this same code already? We have service objects following the command pattern, the same ones being called from here and from the canaries.

So why do we need to make the actual flaggy command exportable? Does the new code that is supposed to import this have new requirements we didn't cover before?

PR #583, opened 2025-06-03
uses proxy environment for connection to host test

jaxesn:
*Issue #, if available:*

*Description of changes:*

Currently during the debug flow, and soon during init, the code which attempts to verify if a host is reachable/connectable is not proxy aware.  We use the `net.Dialer` to try and open a tcp connection then immediately close it.  In the case that the environment requires a proxy for internet access, ex: ssm/iam-ra endpoints, this will fail.

This modifies this check to be proxy aware.  If the url is configured, by the user, to require a proxy, a CONNECT request is made to the proxy with the host as the target url. This will initiate the tunnel thru the proxy to the target. If this succeeds, we immediately close the tunnel and mark the check as success. If this fails, it is assumed the target url is not reachable.

I am working on changes to the e2e stack to support running tests/creating nodes which require a proxy. This will come in a future PR, but that is what I am using to confirm the current and new behavior.

*Testing (if applicable):*

### DIFF
--- a/internal/network/connection.go
+++ b/internal/network/connection.go
@@ -1,30 +1,105 @@
 package network
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"net"
+	"net/http"
 	"net/url"
 	"time"
+
+	"golang.org/x/net/http/httpproxy"
 )
 
 const dialTimeout = 5 * time.Second
 
+type ConnectionOptions struct {
+	// ProxyFunc determines which proxy to use for a given request.
+	// If nil, uses httpproxy.FromEnvironment().ProxyFunc()
+	ProxyFunc func(*url.URL) (*url.URL, error)
+}
+
+type ConnectionOption func(*ConnectionOptions)
+
+// this is overridable mainly for testing purposes to avoid using environment variables
+// to manipulate the proxy environment and because the default ProxyFunc always skips the proxy
+// for localhost/loopback addresses, it does not require that they be listed in the NO_PROXY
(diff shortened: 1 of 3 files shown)


internal/network/connection.go
```diff
@@ -1,30 +1,108 @@
...
+	// ex: HTTP/1.1 200 Connection Established
+	resp, err := http.ReadResponse(bufio.NewReader(conn), req)
+	if err != nil {
+		conn.Close()
+		return fmt.Errorf("proxy: reading CONNECT response: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		conn.Close()
```
g-gaston, line 103: why do we close the connection on all error branches by we don't when the function succeeded?

internal/network/connection_test.go
```diff
@@ -0,0 +1,241 @@
+package network
```
g-gaston, line 1:
```suggestion
package network_test
```

internal/network/connection_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+func TestCheckConnectionToHost(t *testing.T) {
+	tests := []struct {
+		name           string
+		useProxy       bool
+		targetServer   serverFunc
+		proxyBehavior  proxyBehavior
+		proxyShouldErr bool
+		wantErr        bool
+		errContains    string // optional substring to check in error message
```
g-gaston, line 129: these two seem redundant
When we expect an error, we always want to check part of the error message

g-gaston (approved):

PR #590, opened 2025-06-06
adds support to e2e framework to create nodes requiring proxy

jaxesn:
## Description of changes

This PR adds support for creating nodes requiring proxy in the E2E framework, implementing the proxy configuration requirements as documented in the [AWS EKS hybrid nodes proxy documentation](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-proxy.html).

### Key changes
- Implemented proxy configuration for container runtime operations
- Added proxy environment variable handling for:
  - Package manager operations (apt/yum)
  - Container runtime operations
  - System-level proxy settings
- Added proxy support for package manager operations (apt/yum) during node creation
- Added squid to the jumpbox instance to be used as the http_proxy
- Added new ProxyOnly sg for nodes we want to test with limited internet access. The CP cidr is allowed along with the jumpboxinstance.  I intend to update the docs to call out that we strongly recommend that nodes have direct access/routability to the CP private IPs otherwise a number of deployments (ex: cilium) will need the proxy config.  In a public/private cluster, in general kube-proxy running on-prem is going to resolve the kube api endpoint as the public IPs.  It is then going to connect via these public IPs to get the Services to then create the iptable rules.  I want to call this out in the docs as well.  These iptable rules will point to the private IPs of the CP instances and as long as the node has access to these without the need of the proxy, then other deployments do not need the proxy configuration.
- pod identity is a good example of a deployment that will need to proxy config since it is access the AWS api, so would any pods which consume the pod identity metadata endpoint.  pod identity is also unique in that it listens on the metadata like endpoint and expects that not to go thru the proxy, so this needs to be set in the NO_PROXY

### Testing
- Added E2E tests for proxy-enabled node creation scenarios
- Verified package manager operations with proxy settings for both apt and yum-based systems
- Tested container runtime operations through proxy
- Validated on supported operating systems
- 

### DIFF
--- a/test/e2e/cluster/cfn-templates/setup-cfn.yaml
+++ b/test/e2e/cluster/cfn-templates/setup-cfn.yaml
@@ -334,6 +334,69 @@ Resources:
                   - "|"
                   - !Ref PortWithDescription
 
+  HybridNodeDefaultSecurityGroupIngressFromProxyOnly:
+    Type: AWS::EC2::SecurityGroupIngress
+    Properties:
+      GroupId: !GetAtt HybridNodeVPC.DefaultSecurityGroup
+      IpProtocol: tcp
+      FromPort: 3128
+      ToPort: 3128
+      SourceSecurityGroupId: !Ref HybridNodeProxyOnlySecurityGroup
+
+  HybridNodeProxyOnlySecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Security group for hybrid nodes that only allows traffic through the proxy
+      VpcId: !Ref HybridNodeVPC
+      SecurityGroupIngress:
+        - IpProtocol: -1
+          FromPort: -1
+          ToPort: -1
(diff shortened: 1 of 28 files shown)


test/e2e/cluster/cfn-templates/setup-cfn.yaml
```diff
@@ -330,6 +330,69 @@ Resources:
...
+    Type: AWS::EC2::SecurityGroupEgress
+    Properties:
+      GroupId: !Ref HybridNodeProxyOnlySecurityGroup
+      IpProtocol: tcp
+      FromPort: 3128
+      ToPort: 3128
+      DestinationSecurityGroupId: !GetAtt HybridNodeVPC.DefaultSecurityGroup
+
+  HybridNodeProxyOnlySecurityGroupEgressToCluster:
+    Type: AWS::EC2::SecurityGroupEgress
```
g-gaston, line 388: do we need to create a default egress block everything else? i thought the default is egress to everything

g-gaston (approved):

PR #616, opened 2025-06-12
Move supported versions to a constants file for integration tests

achen1210:
*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/test/integration/cases/init-with-config-enrichment/run.sh
+++ b/test/integration/cases/init-with-config-enrichment/run.sh
@@ -5,6 +5,7 @@ set -o nounset
 set -o pipefail
 
 source /helpers.sh
+source /test-constants.sh
 
 mock::aws
 wait::dbus-ready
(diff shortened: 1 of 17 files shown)


test/integration/run.sh
```diff
@@ -34,7 +34,7 @@ for CASE_DIR in $(find test/integration/cases -maxdepth 1 -mindepth 1 -name "$TE
     -v $PWD/$CASE_DIR:/test-case \
```
abhay-krishna, line 34: Since we're only mounting the specific test case folder here and not the parent folder (`test/integration`), the `helpers.sh` and `test-constants.sh` script in the parent folder won't be available on the container.

achen1210: I think the issue actually was that `./run.sh` runs the script in a subshell so it didn't have the env from the current shell. Using `source` instead fixes the issue since it's run in the current shell, but we lose some isolation. I think it's ok like this since we're using a fresh container for each test and only using the container for the test, but I'm not sure this is worth it just to take out the sourcing of helpers and constants for each test case. What do you think?

PR #630, opened 2025-06-28
Add manifest infrastructure for region configuration

tatlat:
- Add RegionConfig and RegionData structures to manifest
- Add GetRegionConfig function to retrieve region-specific configuration
- Add test data and tests for manifest region configuration
- Foundation for reading ECR account IDs from manifest

*Issue #, if available:*

*Description of changes:*

*Testing (if applicable):*

### DIFF
--- a/internal/aws/source.go
+++ b/internal/aws/source.go
@@ -131,6 +131,20 @@ func getLatestDateEksPatchRelease(patchReleases []EksPatchRelease) (EksPatchRele
 	return latestRelease, nil
 }
 
+func GetRegionConfig(ctx context.Context, region string) (*RegionData, error) {
+	manifest, err := getReleaseManifest(ctx)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/aws/source.go
```diff
@@ -131,6 +131,20 @@ func getLatestDateEksPatchRelease(patchReleases []EksPatchRelease) (EksPatchRele
 	return latestRelease, nil
 }
 
+func GetRegionConfig(ctx context.Context, region string) (*RegionData, error) {
+	manifest, err := getReleaseManifest(ctx)
```
achen1210, line 135: We also make this http call to grab the manifest in GetLatestSource so could we combine them and make the call only once during the upgrade flow by saving the manifest or region details in the upgrader struct?

achen1210 (approved):

PR #632, opened 2025-06-28
Use manifest to determine ECR

tatlat:
*Issue #, if available:*

*Description of changes:*
Removed hardcoded ecr region to account mappings for regions where hybrid nodes is supported.
Restructured the logic to use the ecr region account mapping from the manifest instead.

*Testing (if applicable):*

### DIFF
--- a/internal/flows/init.go
+++ b/internal/flows/init.go
@@ -6,6 +6,8 @@ import (
 	"go.uber.org/zap"
 	"k8s.io/utils/strings/slices"
 
+	"github.com/aws/eks-hybrid/internal/aws"
+	"github.com/aws/eks-hybrid/internal/configenricher"
 	"github.com/aws/eks-hybrid/internal/nodeprovider"
 )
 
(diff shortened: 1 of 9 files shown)


internal/flows/init.go
```diff
@@ -33,7 +35,14 @@ func (i *Initer) Run(ctx context.Context) error {
 		return err
 	}
 
-	if err := i.NodeProvider.Enrich(ctx); err != nil {
+	// Get region config from manifest for ECR registry lookup
+	region := i.NodeProvider.GetNodeConfig().Spec.Cluster.Region
+	regionConfig, err := aws.GetRegionConfig(ctx, region)
+	if err != nil {
+		i.Logger.Warn("Failed to get region config from manifest, using fallback ECR registry logic", zap.Error(err))
```
achen1210, line 42: nit: maybe take out ecr fallback reference here since logging is in `ecr.go`

PR #633, opened 2025-06-30
Add option to skip pod identity tests via SKIP_POD_IDENTITY_TEST

achen1210:
*Issue #, if available:*

*Description of changes:*

Adds a skip option for pod identity tests using SKIP_POD_IDENTITY_TEST akin to the flag SKIP_IRA_TEST for iam-ra tests. This will be used to test new versions of EKS before the pod identity add-on is available.

*Testing (if applicable):*

### DIFF
--- a/test/e2e/cluster/create.go
+++ b/test/e2e/cluster/create.go
@@ -152,17 +152,19 @@ func (c *Create) Run(ctx context.Context, test TestResources) error {
 		return fmt.Errorf("creating kubernetes client: %w", err)
 	}
 
-	podIdentityAddon := addon.NewPodIdentityAddon(hybridCluster.Name, stackOut.podIdentity.roleArn)
+	if !skipPodIdentityTest() {
+		podIdentityAddon := addon.NewPodIdentityAddon(hybridCluster.Name, stackOut.podIdentity.roleArn)
 
-	err = podIdentityAddon.Create(ctx, c.logger, c.eks, k8sClient)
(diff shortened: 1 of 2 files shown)


test/e2e/cluster/create.go
```diff
@@ -152,17 +152,19 @@ func (c *Create) Run(ctx context.Context, test TestResources) error {
 		return fmt.Errorf("creating kubernetes client: %w", err)
 	}
 
-	podIdentityAddon := addon.NewPodIdentityAddon(hybridCluster.Name, stackOut.podIdentity.roleArn)
+	if !skipPodIdentityTest() {
```
junpengdev, line 155: why do we need to skip pod identity add-on test but not other add-on?

junpengdev: do they build a new pod identity add-on for every new kubernetes version?

achen1210: Yes, the pod identity add-ons are tied to specific kubernetes versions, they're usually not available right away when testing new pre-GA versions. The other add-ons are running in separate test suite, but we want to verify the core test suite earlier in the release testing process so we don't want pod identity to be the only reason those are failing, hence the option to skip it

achen1210: I think it will currently will error out, this skip would allow us to test just the base hybrid cluster functionality without any add-ons during pre-release testing

PR #635, opened 2025-06-30
Add conditional logic to set endpoints for beta

achen1210:
*Issue #, if available:*

*Description of changes:*
Sets the endpoints to Beta endpoints if the environment is beta - to be used for new version testing pipeline
*Testing (if applicable):*

### DIFF
--- a/buildspecs/test-nodeadm.yml
+++ b/buildspecs/test-nodeadm.yml
@@ -9,6 +9,12 @@ phases:
   pre_build:
     commands:
       - echo "Setting up for e2e tests..."
+      - |
+        if [ "$ENVIRONMENT" = "beta" ]; then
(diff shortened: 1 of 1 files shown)


buildspecs/test-nodeadm.yml
```diff
@@ -9,6 +9,12 @@ phases:
   pre_build:
     commands:
       - echo "Setting up for e2e tests..."
+      - |
```
jaxesn, line 12: we have the beta job that already sets these? is this so that we can run against beta in the regular ci run?

achen1210: Yeah it's for the single version testing pipeline for doing pre GA testing of new EKS versions - wanted to keep it separate from the normal pipelines in case the new version has issues or some components aren't ready yet and cause failures during testing

PR #643, opened 2025-07-08
Adding Mixed-Mode tests to e2e

ppadyala1599:
Adding end-to-end testing for mixed mode EKS clusters where hybrid nodes and cloud-managed nodes coexist. 

Test Coverage :

- Cross-VPC pod-to-pod communication (both directions)

- Cross-VPC service discovery with DNS resolution across node types (both directions)

- CoreDNS optimization for mixed mode

### DIFF
--- a/test/e2e/kubernetes/pod.go
+++ b/test/e2e/kubernetes/pod.go
@@ -31,12 +31,32 @@ func GetNginxPodName(name string) string {
 	return "nginx-" + name
 }
 
-func CreateNginxPodInNode(ctx context.Context, k8s kubernetes.Interface, nodeName, namespace, region string, logger logr.Logger) error {
-	podName := GetNginxPodName(nodeName)
+func CreateNginxPodInNode(ctx context.Context, k8s kubernetes.Interface, nodeName, namespace, region string, logger logr.Logger, options ...interface{}) error {
+	var podName string
+	var labels map[string]string
+
+	// Parse optional parameters
+	for _, option := range options {
+		switch v := option.(type) {
+		case string:
+			if v != "" {
+				podName = v
+			}
+		case map[string]string:
+			labels = v
+		}
+	}
+
+	// Use default pod name if not provided
+	if podName == "" {
+		podName = GetNginxPodName(nodeName)
+	}
+
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      podName,
 			Namespace: namespace,
+			Labels:    labels,
 		},
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{
@@ -61,7 +81,6 @@ func CreateNginxPodInNode(ctx context.Context, k8s kubernetes.Interface, nodeNam
 					},
 				},
 			},
-			// schedule the pod on the specific node using nodeSelector
 			NodeSelector: map[string]string{
 				"kubernetes.io/hostname": nodeName,
 			},
@@ -271,14 +290,14 @@ func TestPodToPodConnectivity(ctx context.Context, config *restclient.Config, k8
 	targetIP := targetPod.Status.PodIP
 	logger.Info("Testing pod-to-pod connectivity", "from", clientPodName, "to", targetPodName, "ip", targetIP)
 
-	cmd := []string{"curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", fmt.Sprintf("http://%s:8080", targetIP), "--connect-timeout", "10", "--max-time", "30"}
+	cmd := []string{"curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", fmt.Sprintf("http://%s:80", targetIP), "--connect-timeout", "10", "--max-time", "30"}
 
 	_, err = ik8s.GetAndWait(ctx, 2*time.Minute, k8s.CoreV1().Pods(namespace), clientPodName, func(pod *corev1.Pod) bool {
 		if pod.Status.Phase != corev1.PodRunning {
 			return false
 		}
 
-		result, execErr := ExecInPod(ctx, config, k8s, clientPodName, namespace, cmd)
+		result, _, execErr := execPod(ctx, config, k8s, clientPodName, namespace, cmd...)
 		if execErr != nil {
 			logger.Info("Pod connectivity test failed", "error", execErr.Error())
 			return false
@@ -294,16 +313,43 @@ func TestPodToPodConnectivity(ctx context.Context, config *restclient.Config, k8
 	return nil
 }
 
-// ExecInPod executes a command in a pod and returns stdout (simplified version of ExecPodWithRetries)
-func ExecInPod(ctx context.Context, config *restclient.Config, k8s kubernetes.Interface, podName, namespace string, command []string) (string, error) {
-	stdout, _, err := execPod(ctx, config, k8s, podName, namespace, command...)
-	return stdout, err
-}
-
 func WaitForDaemonSetPodToBeRunning(ctx context.Context, k8s kubernetes.Interface, namespace, daemonSetName, nodeName string, logger logr.Logger) error {
 	listOptions := metav1.ListOptions{
 		LabelSelector: fmt.Sprintf("%s=%s", "app.kubernetes.io/name", daemonSetName),
 		FieldSelector: fmt.Sprintf("%s=%s", "spec.nodeName", nodeName),
 	}
 	return WaitForPodsToBeRunning(ctx, k8s, listOptions, namespace, logger)
+}
+
+// ListPodsWithLabels lists pods in a namespace that match the given label selector
+func ListPodsWithLabels(ctx context.Context, k8s kubernetes.Interface, namespace, labelSelector string) (*corev1.PodList, error) {
+	pods, err := k8s.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labelSelector,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list pods with selector %s: %w", labelSelector, err)
+	}
+	return pods, nil
+}
+
+// DeletePodsWithLabels deletes all pods in a namespace that match the given label selector
+func DeletePodsWithLabels(ctx context.Context, k8s kubernetes.Interface, namespace, labelSelector string, logger logr.Logger) error {
+	pods, err := ListPodsWithLabels(ctx, k8s, namespace, labelSelector)
+	if err != nil {
+		return fmt.Errorf("failed to list pods for deletion: %w", err)
+	}
+
+	for _, pod := range pods.Items {
+		// Force delete pods to prevent resource accumulation issues
+		gracePeriodSeconds := int64(0)
+		deleteOptions := metav1.DeleteOptions{
+			GracePeriodSeconds: &gracePeriodSeconds,
+		}
+		if err := k8s.CoreV1().Pods(namespace).Delete(ctx, pod.Name, deleteOptions); err != nil {
+			logger.Info("Pod cleanup: resource not found or already deleted", "name", pod.Name)
+		} else {
+			logger.Info("Force deleted pod", "name", pod.Name)
+		}
+	}
+	return nil
 }
--- a/test/e2e/kubernetes/pod.go
+++ b/test/e2e/kubernetes/pod.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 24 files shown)


test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,643 @@
...
+
+var _ = SynchronizedBeforeSuite(
+	// This function only runs once, on the first process
+	func(ctx context.Context) []byte {
+		suiteConfig := suite.BeforeSuiteCredentialSetup(ctx, filePath)
+
+		test := suite.BeforeVPCTest(ctx, &suiteConfig)
+		credentialProviders := suite.AddClientsToCredentialProviders(suite.CredentialProviders(), test)
+		osProviderList := suite.OSProviderList(credentialProviders)
+		randomOSProvider := osProviderList[rand.Intn(len(osProviderList))]
```
2ez4szliu, line 60: does this mean that in one e2e test run, we pick a random OS & cred provider to run the mixed mode connectivity test?

test/e2e/suite/peered_vpc.go
```diff
@@ -527,3 +544,101 @@ func CreateNodes(ctx context.Context, test *PeeredVPCTest, nodesToCreate []NodeC
...
+		return fmt.Errorf("no public subnets found for managed node groups")
+	}
+
+	input := &eks.CreateNodegroupInput{
+		ClusterName:   aws.String(t.Cluster.Name),
+		NodegroupName: aws.String(nodeGroupName),
+		Subnets:       validSubnets,
+		NodeRole:      aws.String(t.StackOut.ManagedNodeRoleArn),
+		InstanceTypes: []string{"m5.large"},
+		AmiType:       ekstypes.AMITypesAl2X8664,
```
abhay-krishna, line 580: Is there any reason we're not using AL2023 here? It is the latest stable version and the docs even recommend using it.

ppadyala1599: I updated this to `Al2023X8664Standard`

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,643 @@
...
+				testServiceConnectivityWithRetries(ctx, test, clientPod.Name, service.Name, 8080)
+
+				test.Logger.Info("Bidirectional service discovery test (cloud → hybrid) completed successfully")
+			})
+		})
+	})
+})
+
+// createPodWithSelector creates a pod with the specified configuration
+func createPodWithSelector(ctx context.Context, test *suite.PeeredVPCTest, name, image string, nodeSelector map[string]string, port int32) *corev1.Pod {
```
abhay-krishna, line 174: Why not use the `CreatePod` method already available in the `kubernetes` package? It also has helper methods for idempotent creates and waits.

test/e2e/kubernetes/pod.go
```diff
@@ -256,6 +257,49 @@ func execPod(ctx context.Context, config *restclient.Config, k8s kubernetes.Inte
...
+	if err != nil {
+		return fmt.Errorf("connectivity from %s to %s failed: %w", clientPodName, targetPodName, err)
+	}
+
+	logger.Info("Pod-to-pod connectivity successful", "from", clientPodName, "to", targetPodName)
+	return nil
+}
+
+// ExecInPod executes a command in a pod and returns stdout (simplified version of ExecPodWithRetries)
+func ExecInPod(ctx context.Context, config *restclient.Config, k8s kubernetes.Interface, podName, namespace string, command []string) (string, error) {
```
achen1210, line 298: Since this wrapper function doesn't add anything maybe we can just make `execPod` exported and remove this?

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,408 @@
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "nginx-cloud",
+						Namespace: "default",
+						Labels:    map[string]string{"app": "nginx-cloud"},
+					},
+					Spec: corev1.PodSpec{
+						NodeSelector: cloudNodeSelector,
+						Containers: []corev1.Container{{
+							Name:    "nginx-cloud",
+							Image:   "nginx:1.21",
```
abhay-krishna, line 118: Can we use the nginx image used in the nginx pod helper method [here](https://github.com/aws/eks-hybrid/blob/main/test/e2e/kubernetes/pod.go#L44)? This helps to avoid the Dockerhub dependency and unauthenticated pull rate limits. Also if you can find a way to reuse that method instead of creating a new pod object from scratch, that would be great.

Also the 1.21 version seems to be pretty old, ~3 years. Why are we using that version? Can we bump to newer versions? Or as suggested above, you could use the image from `CreateNginxPodInNode` method which gives you the latest one.

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,408 @@
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "nginx-hybrid-reverse",
+						Namespace: "default",
+						Labels:    map[string]string{"app": "nginx-hybrid-reverse"},
+					},
+					Spec: corev1.PodSpec{
+						NodeSelector: hybridNodeSelector,
+						Containers: []corev1.Container{{
+							Name:    "nginx-hybrid-reverse",
+							Image:   "nginx:1.21",
```
abhay-krishna, line 165: Same comment as above regarding the nginx image.

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,408 @@
...
+				err = kubernetes.TestPodToPodConnectivity(ctx, test.K8sClientConfig, test.K8sClient(), clientPod.Name, hybridPod.Name, "default", test.Logger)
+				Expect(err).NotTo(HaveOccurred(), "testing pod-to-pod connectivity")
+
+				test.Logger.Info("Cross-VPC pod-to-pod communication test completed successfully")
+			})
+		})
+
+		Context("Cross-VPC Service Discovery", func() {
+			It("should enable cross-VPC service discovery from hybrid to cloud services", func(ctx context.Context) {
+				service, _, err := kubernetes.CreateServiceWithDeployment(ctx, test.K8sClient(), "nginx-service-cloud", "nginx:1.21", cloudNodeSelector, 80, 80, 1, test.Logger)
```
abhay-krishna, line 203: The 1.21 version seems to be pretty old, ~3 years. Why are we using that version? Can we bump to newer versions? Or as suggested above, you could use the [image](https://github.com/aws/eks-hybrid/blob/main/test/e2e/kubernetes/pod.go#L44) from `CreateNginxPodInNode` method which gives you the latest one.

test/e2e/kubernetes/service.go
```diff
@@ -0,0 +1,179 @@
...
+// WaitForServiceReady waits for service endpoints to be ready
+func WaitForServiceReady(ctx context.Context, k8s kubernetes.Interface, serviceName, namespace string, logger logr.Logger) error {
+	logger.Info("Waiting for service endpoints", "service", serviceName)
+
+	_, err := ik8s.ListAndWait(ctx, 3*time.Minute, k8s.DiscoveryV1().EndpointSlices(namespace), func(endpointSlices *discoveryv1.EndpointSliceList) bool {
+		for _, slice := range endpointSlices.Items {
+			if ownerService, exists := slice.Labels["kubernetes.io/service-name"]; exists && ownerService == serviceName {
+				for _, endpoint := range slice.Endpoints {
+					if endpoint.Conditions.Ready != nil && *endpoint.Conditions.Ready {
+						return true
```
achen1210, line 111: Looks like our services we make only have 1 endpoint so this is ok but maybe its better to make sure all endpoints are ready in case we change this in the future? i.e. flip to `endpoint.Conditions.Ready == nil || !*endpoint.Conditions.Ready -> return false` and then return true at end of loop

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,408 @@
...
+	RunSpecs(t, "Mixed Mode E2E Suite")
+}
+
+var _ = SynchronizedBeforeSuite(
+	// This function only runs once, on the first process
+	func(ctx context.Context) []byte {
+		suiteConfig := suite.BeforeSuiteCredentialSetup(ctx, filePath)
+
+		test := suite.BeforeVPCTest(ctx, &suiteConfig)
+		credentialProviders := suite.AddClientsToCredentialProviders(suite.CredentialProviders(), test)
```
achen1210, line 55: Would it be better to fix the provider/OS or just test multiple combinations consistently? I think most of these tests should by OS agnostic and randomizing amongst different OS versions during each test run might make it difficult to isolate any test infrastructure failures related to the OS - if we want to cover a few different ones we could also have some kind of consistent schedule maybe

achen1210:
We might want to add managed node group cleanup to the sweeper also: https://github.com/aws/eks-hybrid/blob/bcef70587ce747f7fcd931419c92c67066ba7aef/test/e2e/cleanup/sweeper.go

test/e2e/kubernetes/pod.go
```diff
@@ -256,6 +283,43 @@ func execPod(ctx context.Context, config *restclient.Config, k8s kubernetes.Inte
...
+	logger.Info("Testing pod-to-pod connectivity", "from", clientPodName, "to", targetPodName, "ip", targetIP)
+
+	cmd := []string{"curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", fmt.Sprintf("http://%s:80", targetIP), "--connect-timeout", "10", "--max-time", "30"}
+
+	_, err = ik8s.GetAndWait(ctx, 2*time.Minute, k8s.CoreV1().Pods(namespace), clientPodName, func(pod *corev1.Pod) bool {
+		if pod.Status.Phase != corev1.PodRunning {
+			return false
+		}
+
+		result, _, execErr := ExecPodWithRetries(ctx, config, k8s, clientPodName, namespace, cmd...)
```
achen1210, line 307: Do we want this to retry also or just use the retries in the GetAndWait for the exec into the pod as well?

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	rules := []struct {
+		protocol string
+		port     int32
+		cidr     string
+		desc     string
+	}{
+		// HTTP connectivity for pod-to-pod and service tests
+		{"tcp", 80, "10.0.0.0/16", "HTTP 80 from cloud VPC"},
```
abhay-krishna, line 281: Let's change these to match the defaults [here](https://github.com/aws/eks-hybrid/blob/ce6ef2f2fff19783edd08b56fc1dc7e87a50ce0c/test/e2e/cluster/create.go#L240).

test/e2e/suite/peered_vpc.go
```diff
@@ -262,6 +264,21 @@ func (t *PeeredVPCTest) NewVerifyPodIdentityAddon(nodeName string) *addon.Verify
 	}
 }
 
+// K8sClient returns the Kubernetes client for external package access
```
abhay-krishna, line 267: What's the tradeoff in using getters vs exporting the fields? Since this is just tests, I'm not concerned with exporting the fields, in fact, I vote for doing that. Getters/setters have their place in Go but I don't think this is one of them.

test/e2e/kubernetes/pod.go
```diff
@@ -30,12 +31,21 @@ func GetNginxPodName(name string) string {
...
+	} else {
+		podName = GetNginxPodName(nodeName)
+	}
+
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      podName,
 			Namespace: namespace,
+			Labels: map[string]string{
+				"test-suite": "e2e-test",
```
achen1210, line 47: This could be a parameter too I think

test/e2e/suite/mixed_mode/mixed_mode_test.go
ppadyala1599: i think even mixed mode is designed to be OS-agnostic, implementing randomized testing across different providers and operating systems would help us ensure comprehensive coverage and make sure it works. This could help us identify potential issues early if any and ensure relaibility.

test/e2e/suite/mixed_mode/mixed_mode_test.go
ppadyala1599: I refactored all pod creation to use existing kubernetes.CreateNginxPodInNode() method by adding optional parameters for custom pod name and labels for mixed mode testing

test/e2e/suite/mixed_mode/mixed_mode_test.go
ppadyala1599: refactored all pod creation to use existing kubernetes.CreateNginxPodInNode() method by adding optional parameters for custom pod name and labels

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,408 @@
...
+				// Test service connectivity with integrated DNS resolution testing
+				err = kubernetes.TestServiceConnectivityWithRetries(ctx, test.K8sClientConfig, test.K8sClient(), clientPod.Name, service.Name, "default", 8080, test.Logger)
+				Expect(err).NotTo(HaveOccurred(), "testing service connectivity")
+
+				test.Logger.Info("Cross-VPC service discovery test (hybrid → cloud) completed successfully")
+			})
+
+			It("should enable cross-VPC service discovery from cloud to hybrid services", func(ctx context.Context) {
+				test.Logger.Info("Testing bidirectional service discovery (cloud → hybrid service)")
+				service, _, err := kubernetes.CreateServiceWithDeployment(ctx, test.K8sClient(), "nginx-service-hybrid", "nginx:1.21", hybridNodeSelector, 80, 80, 1, test.Logger)
```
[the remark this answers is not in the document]
ppadyala1599: refactored all pod creation to use existing kubernetes.CreateNginxPodInNode() method by adding optional parameters for custom pod name and labels

test/e2e/suite/mixed_mode/mixed_mode_test.go
ppadyala1599: https://github.com/aws/eks-hybrid/blob/main/test/e2e/README.md?plain=1#L86C1-L103

test/e2e/kubernetes/pod.go
```diff
@@ -71,6 +91,24 @@ func CreateNginxPodInNode(ctx context.Context, k8s kubernetes.Interface, nodeNam
 	return CreatePod(ctx, k8s, pod, logger)
 }
 
+// FindNodeWithLabel finds a node that has the specified label key-value pair
+func FindNodeWithLabel(ctx context.Context, k8s kubernetes.Interface, labelKey, labelValue string, logger logr.Logger) (string, error) {
```
abhay-krishna, line 95: This should be in node.go

test/e2e/suite/mixed_mode/mixed_mode_test.go
```diff
@@ -0,0 +1,541 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to configure CoreDNS preferred anti-affinity: %w", err)
+	}
+
+	// Step 3: Configure kube-dns service for traffic distribution
+	serviceTrafficPatch := `{
+		"spec": {
+			"trafficDistribution": "PreferClose"
+		}
+	}`
```
abhay-krishna, line 494:
```suggestion
	"spec": {
		"trafficDistribution": "PreferClose"
	}
}`
```

abhay-krishna (approved):

PR #669, opened 2025-07-25
Remove 1.27 and its feature gates from upgrade & kubelet config tests

achen1210:
*Issue #, if available:*

*Description of changes:*

Removes k8s 1.27 from the integration tests to unblock other PRs. This includes removing the `KubeletCredentialProviders` feature gate from the kubelet config to which became default behavior in 1.28. A second PR will clean up old tests and feature gate support that is now default behavior in all supported Kubernetes versions.

*Testing (if applicable):*

### DIFF
--- a/test/integration/cases/upgrade-with-iam/expected-kubelet-config-initial
+++ b/test/integration/cases/upgrade-with-iam/expected-kubelet-config-initial
@@ -32,7 +32,6 @@
     "nodefs.inodesFree": "5%"
   },
   "featureGates": {
-    "KubeletCredentialProviders": true,
     "RotateKubeletServerCertificate": true
   },
   "hairpinMode": "hairpin-veth",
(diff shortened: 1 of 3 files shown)


tatlat (approved):

test/integration/cases/upgrade-with-iam/expected-kubelet-config-initial
```diff
@@ -32,7 +32,6 @@
     "nodefs.inodesFree": "5%"
   },
   "featureGates": {
-    "KubeletCredentialProviders": true,
```
abhay-krishna, line 35: With the removal of the 1.27 feature gates, the initial and upgraded kubelet configs have become identical. So I don't think we need two separate files now to compare before/after of upgrade. I think we can even remove the separate checks in the run.sh for these cases and maybe have only one after init.

achen1210: Will take it out in the next PR. We have a couple of tests that need cleaning up and/or deleting as well as some dead code [here](https://github.com/aws/eks-hybrid/blob/main/internal/kubelet/config.go#L232) now that only 1.28+ is supported.

PR #698, opened 2025-08-12
Validate cloudwatch observability agent addon  in mixed mode

ppadyala1599:
Added timeout to prevent from hanging indefinitely and reaching Global timeout when waiting for addon to reach active/degraded state.

Validating functionality of Cloud Watch Observability addon in mixed mode environment.

### DIFF
--- a/hybrid-nodes-cdk/lib/nodeadm/policies.ts
+++ b/hybrid-nodes-cdk/lib/nodeadm/policies.ts
@@ -7,7 +7,7 @@ export function createNodeadmTestsCreationCleanupPolicy(
   testClusterPrefix: string,
   binaryBucketArn: string,
   podIdentityS3BucketPrefix: string,
-) {
+): iam.ManagedPolicy[] {
   const requestTagCondition = {
     StringLike: {
       [`aws:RequestTag/${testClusterTagKey}`]: `${testClusterPrefix}-*`,
@@ -18,7 +18,9 @@ export function createNodeadmTestsCreationCleanupPolicy(
       [`aws:ResourceTag/${testClusterTagKey}`]: `${testClusterPrefix}-*`,
     },
   };
-  return new iam.Policy(stack, 'nodeadm-e2e-tests-runner-policy', {
+
+ 
+  const iamPolicy = new iam.ManagedPolicy(stack, 'nodeadm-e2e-tests-iam-policy', {
     statements: [
       new iam.PolicyStatement({
         effect: iam.Effect.ALLOW,
@@ -77,14 +79,57 @@ export function createNodeadmTestsCreationCleanupPolicy(
         conditions: requestTagCondition,
       }),
       new iam.PolicyStatement({
+        effect: iam.Effect.ALLOW,
         actions: [
-          'cloudformation:ListStacks',
-          'cloudformation:DescribeStacks',
-          'ec2:AcceptVpcPeeringConnection',
-          'ec2:AssociateRouteTable',
-          'ec2:AssociateTransitGatewayRouteTable',
-          'ec2:AttachInternetGateway',
-          'ec2:AuthorizeSecurityGroupIngress',
+          'iam:ListOpenIDConnectProviders',
+          'iam:CreateOpenIDConnectProvider',
+          'iam:UpdateAssumeRolePolicy',
+        ],
+        resources: [
+          `arn:aws:iam::${stack.account}:oidc-provider/*`,
+          `arn:aws:iam::${stack.account}:role/*CloudWatch*`,
+        ],
+      }),
+    ],
+  });
+
+  const ec2Policy = new iam.ManagedPolicy(stack, 'nodeadm-e2e-tests-ec2-policy', {
+    statements: [
+      new iam.PolicyStatement({
+        actions: [
+          'ec2:DescribeInstances',
+          'ec2:DescribeInstanceTypes',
+          'ec2:DescribeInstanceAttribute',
+          'ec2:DescribeInstanceStatus',
+          'ec2:DescribeSecurityGroups',
+          'ec2:DescribeSubnets',
+          'ec2:DescribeVpcs',
+          'ec2:DescribeRouteTables',
+          'ec2:DescribeRoutes',
+          'ec2:DescribeTransitGateways',
+          'ec2:DescribeTransitGatewayAttachments',
+          'ec2:DescribeTransitGatewayRouteTables',
+          'ec2:DescribeTransitGatewayVpcAttachments',
+          'ec2:DescribeVpcPeeringConnections',
+          'ec2:DescribeLaunchTemplates',
+          'ec2:DescribeLaunchTemplateVersions',
+          'ec2:DescribeFleets',
+          'ec2:DescribeNetworkInterfaces',
+          'ec2:DescribeSnapshots',
+          'ec2:DescribeImages',
+          'ec2:DescribeKeyPairs',
+          'ec2:DescribeAvailabilityZones',
+          'ec2:DescribeRegions',
+          'ec2:DescribeInternetGateways',
+          'ec2:DescribeNatGateways',
+          'ec2:DescribeVpcEndpoints',
+          'ec2:DescribeTags',
+        ],
+        resources: ['*'],
+        effect: iam.Effect.ALLOW,
+      }),
+      new iam.PolicyStatement({
+        actions: [
           'ec2:CreateFleet',
           'ec2:CreateLaunchTemplate',
           'ec2:CreateLaunchTemplateVersion',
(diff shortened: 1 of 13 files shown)


test/e2e/addon/cloudwatch.go
```diff
@@ -3,135 +3,416 @@ package addon
...
+
+				if err == nil && len(streams.LogStreams) > 0 {
+					logger.Info("Log group has active streams - CloudWatch receiving logs", "logGroup", logGroupName, "streamCount", len(streams.LogStreams))
+				}
+				break
+			}
+		}
+
+		if !found {
+			logger.Info("Log group not found", "logGroup", logGroupName)
```
achen1210, line 340: If the log group isn't there maybe we should have it error? Are there cases it won't be there/is delayed and cause premature errors?

test/e2e/addon/cloudwatch.go
```diff
@@ -3,135 +3,416 @@ package addon
...
 		Addon: Addon{
 			Cluster: cluster,
 			Name:    cloudwatchAddonName,
 		},
+		roleArn: roleArn,
 	}
 }
 
-// WaitForComponents waits for CloudWatch components to be ready
-func (cw CloudWatchAddon) WaitForComponents(ctx context.Context, k8sClient clientgo.Interface, logger logr.Logger) error {
```
achen1210, line 37: Do we want to keep this check for components since now we're only checking for add-on ACTIVE state but we can still continue even if add-on isn't active? This way we could ensure that all the agent pods are in ready state at least

test/e2e/addon/cloudwatch.go
```diff
@@ -3,135 +3,416 @@ package addon
...
+		logger.Info("CloudWatch log groups verification had issues but will continue", "error", err.Error())
+	} else {
+		logger.Info("CloudWatch log groups verification successful ")
+	}
+
+	return nil
+}
+
+// VerifyCloudWatchLogGroups verifies that CloudWatch log groups are created and receiving logs
+func (cw *CloudWatchAddon) VerifyCloudWatchLogGroups(ctx context.Context, cwLogsClient *cloudwatchlogs.Client, logger logr.Logger) error {
```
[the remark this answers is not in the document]
ppadyala1599: i think this is sufficient because it validates everything by confirming all four required log groups exist with active log streams, proving that agents deployed successfully across both cloud and hybrid nodes, established IRSA authentication, and connectivity to CloudWatch APIs. This catches all major failure modes (authentication, network, deployment, and configuration issues)  and comprehensive pod-level verification would add minimal value.

test/e2e/addon/cloudwatch.go
```diff
@@ -3,135 +3,416 @@ package addon
...
+		logger.Info("CloudWatch log groups verification successful - found active log groups", "foundGroups", foundLogGroups, "expectedGroups", len(expectedLogGroups))
+		return nil
+	} else {
+		logger.Info("CloudWatch log groups verification completed - no log groups found yet (this may be expected for new clusters)")
+		return nil
+	}
+}
+
+// testWebhookValidation verifies webhook rejects invalid resource specifications
+func (cw *CloudWatchAddon) testWebhookValidation(ctx context.Context, logger logr.Logger) error {
```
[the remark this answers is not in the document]
ppadyala1599: In that case we might end up unnecessarily testing controller pod placement on different nodes because it only deals with K8s-level operations, making similar API calls to the same K8s API endpoints, and this doesn't differ based on the node type. Also there are no nodeSelectors or tolerations that would restrict the controller pod to only one node type - the scheduler can place it on any node type in multiple runs, and  we are sure that both node types already successfully run identical CloudWatch DaemonSet pods with working service discovery.

test/e2e/kubernetes/node.go
```diff
@@ -395,3 +398,121 @@ func CountCoreDNSDistribution(ctx context.Context, k8s kubernetes.Interface, pod
 
 	return hybridCount, cloudCount
 }
+
+// HandleRHELNodes detects RHEL nodes and creates required directories for CloudWatch
```
junpengdev, line 402: In RHEL, the journal log location is `/run/log/journal`. Creating `/var/log/journal` directory does not mean RHEL would save journal logs here, right?

junpengdev (changes requested):

hybrid-nodes-cdk/lib/nodeadm/policies.ts
```diff
@@ -268,6 +310,50 @@ export function createNodeadmTestsCreationCleanupPolicy(
...
+      new iam.PolicyStatement({
+        actions: ['secretsmanager:GetSecretValue'],
+        resources: [`arn:aws:secretsmanager:${stack.region}:${stack.account}:secret:*`],
+      }),
+    ],
+  });
+
+
+  const additionalPolicy = new iam.ManagedPolicy(stack, 'nodeadm-e2e-tests-additional-policy', {
+    statements: [
```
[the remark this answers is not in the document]
ppadyala1599: yes I tested this on my dev stack

test/e2e/kubernetes/node.go
ppadyala1599: agree, this was meant to be a temporary fix as the fluent bit hybrid pod (with rhel) fails to go to running state due to missing `/var/log/journal` when did `kubectl logs` . but will change to use run/log/journal instead in this case.
working on a different pr to configure the directory during the installation itself

hybrid-nodes-cdk/lib/nodeadm/policies.ts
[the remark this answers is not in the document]
junpengdev: Do you also add new permissions to AWSModelRocketHybridNodesBuildCDK and AwsModelRocketHybridNodesNodeadmCanaryCDK?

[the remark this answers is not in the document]
junpengdev: New permissions need to be deployed to AWSModelRocketHybridNodesBuildCDK before we can merge this PR. Otherwise it would break e2e pipeline immediately.

test/e2e/kubernetes/node.go
```diff
@@ -395,3 +397,78 @@ func CountCoreDNSDistribution(ctx context.Context, k8s kubernetes.Interface, pod
 
 	return hybridCount, cloudCount
 }
+
+// HandleRHELNodes detects RHEL nodes and patches FluentBit ConfigMap for RHEL compatibility
+// Fix for issue where FluentBit pods fail on RHEL nodes due to hardcoded /var/log/journal path.
+// RHEL stores journal data in /run/log/journal, so we patch the ConfigMap to use the correct path.
+func HandleRHELNodes(ctx context.Context, k8sClient kubernetes.Interface, logger logr.Logger) error {
```
junpengdev, line 404: I don't see this get called in this PR. Let me know if I miss anything.

test/e2e/kubernetes/node.go
```diff
@@ -395,3 +397,78 @@ func CountCoreDNSDistribution(ctx context.Context, k8s kubernetes.Interface, pod
...
+
+	if err := patchFluentBitConfigMapForRHEL(ctx, k8sClient, logger); err != nil {
+		return fmt.Errorf("patching FluentBit ConfigMap for RHEL compatibility: %w", err)
+	}
+
+	logger.Info("Successfully patched FluentBit ConfigMap for RHEL compatibility")
+	return nil
+}
+
+// patchFluentBitConfigMapForRHEL patches FluentBit ConfigMap to use /run/log/journal for RHEL compatibility
```
junpengdev, line 436: Is it possible to create symbolic link `/var/log/journal` to `/run/log/journal` when bootstrapping RHEL nodes? If it is feasible, then we don't need to configure it differently for fluentbit.

test/e2e/kubernetes/node.go
ppadyala1599: we call this in SetupCwAddon after creating addon

test/e2e/kubernetes/daemonset.go
```diff
@@ -37,3 +38,39 @@ func DaemonSetWaitForReady(ctx context.Context, logger logr.Logger, k8s kubernet
...
+
+	//  Wait for rollout to complete
+	statusArgs := []string{
+		"rollout", "status",
+		fmt.Sprintf("daemonset/%s", name),
+		"-n", namespace,
+		"--timeout=150s",
+	}
+
+	statusCmd := exec.Command("kubectl", statusArgs...)
```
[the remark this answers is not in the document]
ppadyala1599: sure we can do that by patching deamonset to achieve the same

hybrid-nodes-cdk/lib/nodeadm/policies.ts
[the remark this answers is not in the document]
ppadyala1599: yes i have a cr out for that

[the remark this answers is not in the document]
ppadyala1599: we dont have to add any permissions to `AwsModelRocketHybridNodesNodeadmCanaryCDK` as we only change mixed mode and addon suites which are seperate.

test/e2e/kubernetes/node.go
junpengdev: Is `SetupCwAddon` already merged? I couldn't find the func definition in this PR or in `main` branch.

test/e2e/kubernetes/node.go
```diff
@@ -395,3 +397,78 @@ func CountCoreDNSDistribution(ctx context.Context, k8s kubernetes.Interface, pod
...
+func patchFluentBitConfigMapForRHEL(ctx context.Context, k8sClient kubernetes.Interface, logger logr.Logger) error {
+	logger.Info("Patching FluentBit ConfigMap to use /run/log/journal for RHEL node compatibility")
+
+	maxWaitTime := 3 * time.Minute
+	checkInterval := 5 * time.Second
+
+	err := wait.PollUntilContextTimeout(ctx, checkInterval, maxWaitTime, true, func(ctx context.Context) (bool, error) {
+		checkCmd := exec.Command("kubectl", "get", "configmap", "fluent-bit-config", "-n", "amazon-cloudwatch")
```
junpengdev, line 444: Same comment as the one in `daemonset.go`. If we decided this is the route to go, then let's use Kubernetes API to modify the configmap instead of running kubectl with exec.

test/e2e/kubernetes/node.go
junpengdev: What side effects have you observed? If this simple solution can fix your problem for now, then let's not worry tomorrow's problem, which may not happen.

junpengdev (changes requested):

hybrid-nodes-cdk/lib/nodeadm/policies.ts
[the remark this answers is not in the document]
junpengdev: Are these permissions used during infra setup or node setup? Or are they only used in the actual mix mode tests?

hybrid-nodes-cdk/lib/nodeadm/policies.ts
```diff
@@ -352,6 +352,38 @@ export function createNodeadmTestsCreationCleanupPolicy(
...
+          'iam:DeleteRole',
+          'iam:AttachRolePolicy',
+          'iam:DetachRolePolicy',
+          'iam:GetRole',
+          'iam:PutRolePolicy',
+          'iam:DeleteRolePolicy',
+        ],
+        resources: [`arn:aws:iam::${stack.account}:role/*CloudWatch*`],
+        conditions: requestTagCondition,
+      }),
```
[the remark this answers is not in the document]
ppadyala1599: yes i did

test/e2e/kubernetes/node.go
ppadyala1599: its in line number 463 in `test/e2e/addon/cloudwatch.go` file

test/e2e/kubernetes/node.go
ppadyala1599: The only concern with the symlink approach was wanting to maintain isolation and avoid modifying the actual OS state for what is essentially an addon-specific need. But if that's okay, let's go with the symlink solution.

PR #705, opened 2025-08-14
Remove version toggles for old versions

achen1210:
*Issue #, if available:*

*Description of changes:*

Removes version toggles for versions 1.27 and older that are no longer supported.

Some integration tests for kubelet were failing due to kubelet config merging changing (see https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#kubelet-conf-d). While this was enabled starting in 1.28 we've had it enabled it only starting with 1.29: https://github.com/aws/eks-hybrid/blob/d664a085265b0c35119335b0d57c4f4c98cf4d0b/internal/kubelet/config.go#L50
Since 1.28 reaches EOL soon in September 2025 we'll keep this in for now and remove the old logic then. The `kubelet-config`,`kubelet-config-dir`, and `kubelet-config-set-empty` tests will be updated then.

*Testing (if applicable):*

Ran an E2E test.


internal/kubelet/image-credential-provider.go
```diff
@@ -41,28 +40,20 @@ func (k *kubelet) writeImageCredentialProviderConfig() error {
...
 
-	config, err := generateImageCredentialProviderConfig(k.nodeConfig, ecrCredentialProviderBinPath, k.credentialProviderAwsConfig)
+	credentialProviderConfig, err := generateImageCredentialProviderConfig(k.nodeConfig, ecrCredentialProviderBinPath, k.credentialProviderAwsConfig)
 	if err != nil {
 		return err
 	}
 
 	k.flags["image-credential-provider-bin-dir"] = path.Dir(ecrCredentialProviderBinPath)
-	k.flags["image-credential-provider-config"] = imageCredentialProviderConfigPath
+	k.flags["image-credential-provider-credentialProviderConfig"] = imageCredentialProviderConfigPath
```
junpengdev, line 49: Why do we need to change this flag name? I don't think the flag got changed according to this [doc](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-credential-provider/).

test/integration/cases/kubelet-config-new-instance-type/run.sh
```diff
@@ -9,6 +9,7 @@ source /helpers.sh
 config_path=/tmp/aemm-default-config.json
 cat /etc/aemm-default-config.json | jq '.metadata.values."instance-type" = "mock-type.large" | .dynamic.values."instance-identity-document".instanceType = "mock-type.large"' | tee ${config_path}
 mock::aws ${config_path}
+# TODO: bump kubelet version and switch expected output to include drop-in merge when 1.28 is deprecated
 mock::kubelet 1.28.0
```
junpengdev, line 13: why not replace 1.28 with `${CURRENT_VERSION}` here?

junpengdev (approved):

PR #708, opened 2025-08-19
Upgrade aws-sdk-go/aws to aws-sdk-go-v2/aws

junpengdev:
*Issue #, if available:*
`aws-sdk-go` is deprecated

*Description of changes:*
update `aws-sdk-go/aws` to `aws-sdk-go-v2/aws`.

*Testing (if applicable):*
Tested under personal dev stack.

### DIFF
--- a/internal/api/status.go
+++ b/internal/api/status.go
@@ -6,9 +6,9 @@ import (
 	"io"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
-	"github.com/aws/aws-sdk-go/aws"
 
 	ec2extra "github.com/aws/eks-hybrid/internal/aws/ec2"
 )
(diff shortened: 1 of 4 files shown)


internal/api/status.go
```diff
@@ -57,13 +56,13 @@ func getPrivateDNSName(ec2Client *ec2.Client, instanceID string) (string, error)
...
+	privateDNSName := *out.Reservations[0].Instances[0].PrivateDnsName
 	return privateDNSName, nil
 }
 
 func privateDNSNameAvailable(out *ec2.DescribeInstancesOutput) (bool, error) {
 	if out == nil || len(out.Reservations) != 1 || len(out.Reservations[0].Instances) != 1 {
 		return false, fmt.Errorf("reservation or instance not found")
 	}
-	return aws.StringValue(out.Reservations[0].Instances[0].PrivateDnsName) != "", nil
+	return *out.Reservations[0].Instances[0].PrivateDnsName != "", nil
```
achen1210, line 67: This looks like it could potentially be an issue since function is checking if `privateDnsName` is available so it seems like it could be nil. `aws.StringValue(out.Reservations[0].Instances[0].PrivateDnsName)` will return `""` for nil entry previously (https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringValue), whereas now we would get a runtime dereferencing error using `*`. Do we know this field will exist or should we add a nil check first?

abhay-krishna:
Can you also replace the `github.com/aws/aws-sdk-go/aws/awsutil`  dependency with its V2 counterpart `internal/awsutil`?

abhay-krishna (approved):

PR #715, opened 2025-08-26
Fix e2e test ssh command error for regular OS

junpengdev:
## Issue ##

node `ssh` command fails for Ubuntu, AL, RHEL nodes.

```
❯ ./_bin/e2e-test ssh i-0db5154e378626abd

Starting session with SessionId: jupeng-Isengard-5zo8eisx57lejknqbpdufbr4c8
Warning: Permanently added '10.1.1.138' (ED25519) to the list of known hosts.
ec2-user@10.1.1.138: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).


Exiting session with sessionId: jupeng-Isengard-5zo8eisx57lejknqbpdufbr4c8.
```

The issue was introduced by BottleRocket e2e test, which modifies the ssh command to use `ec2-user` as the user. However, our regular OS (Ubuntu, RHEL, AL), does not have these users when it's provisioned. Thus ssh command failed. 

*Description of changes:*
Now it will check what OS image it is. If it's bottlerocket, then it uses ec2-user for ssh command. Otherwise, it would default to root.

*Testing (if applicable):*
Manually tested

### DIFF
--- a/cmd/e2e-test/ssh/ssh.go
+++ b/cmd/e2e-test/ssh/ssh.go
@@ -82,6 +82,19 @@ func (s *Command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 
 	targetInstance := instances.Reservations[0].Instances[0]
 
+	isBottleRocket, err := isBottleRocket(ctx, ec2Client, *targetInstance.ImageId)
+	if err != nil {
+		return fmt.Errorf("validating if instance OS is BottleRoceket: %w", err)
+	}
+	var sshCommandFormat string
+	if isBottleRocket {
+		sshCommandFormat = "{\"command\":[\"sudo ssh ec2-user@%s\"]}"
+	} else {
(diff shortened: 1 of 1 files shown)


cmd/e2e-test/ssh/ssh.go
```diff
@@ -82,6 +82,19 @@ func (s *Command) Run(log *zap.Logger, opts *cli.GlobalOptions) error {
 
 	targetInstance := instances.Reservations[0].Instances[0]
 
+	isBottleRocket, err := isBottleRocket(ctx, ec2Client, *targetInstance.ImageId)
+	if err != nil {
+		return fmt.Errorf("validate if it is BottleRoceket OS: %w", err)
```
abhay-krishna, line 87:
```suggestion
		return fmt.Errorf("validating if instance OS is Bottlerocket: %w", err)
```

abhay-krishna (approved):

